PR #1, opened 2018-02-26
Added guarded-run to ease into fully automated merges

rousskov:
We need a way to move from staged runs to automated commits without
risking massive wrong automated commits.

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,12 +156,17 @@ request state:
   consult CI logs to determine what happened. The bot does not attempt
   to merge this PR again until the PR branch or PR target branch change.
   When the bot notices that change, it removes this label.
+* `M-cleared-for-merge`: A human allowed the bot running in
+  `config::staged_run` mode to merge the PR. The label has no effect
+  unless the bot is running in that mode. The bot never sets this label.
+  The bot removes this label after successfully merging the PR.
 * `M-merged`: The PR was successfully merged (and probably closed).
   The bot will not attempt to merge this PR again even if it is
   reopened. The bot never removes this label.
 
-All labels, except `M-merged`, are ignored by the bot itself! Humans
-may find them useful when determining the current state of a PR.
+All labels, except `M-cleared-for-merge` and `M-merged`, are ignored by
+the bot itself! Humans may find them useful when determining the current
+state of a PR.
 
 
 ## Commit message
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -271,7 +276,7 @@ All configuration fields are required.
 *repo* | The name of the GitHub repository that the bot should serve. | "squid"
 *owner* | The owner (a person or organization) of the GitHub repository. | "squid-cache"
 *dry_run*| A boolean option to enable read-only, no-modifications mode where the bot logs pull requests selected for merging but skips further merging steps, including PR labeling and commit tagging | false
-*staged_run*| A boolean option to enable no-final-modifications mode where the bot performs all the merging steps up to (and not including) the target branch update. Eligible PRs are merged into and tested on the staging branch but are never merged into their target branches. | false
+*staged_run*| A boolean option instructing the bot to avoid target branch updates unless the PR has a human-set `M-cleared-for-merge` label. When enabled, an eligible PR is merged into and tested on the staging branch. A successfully tested staged commit is then automatically merged into its target branch if and only if the PR has an `M-cleared-for-merge` label. | false
```
eduard-bagdasaryan: Probably we could describe this simpler if we say something like "the bot ignores this option for PRs marked with M-cleared-for-merge label", i.e., the effect the label really performs.

README.md
```diff
@@ -156,12 +156,17 @@ request state:
   consult CI logs to determine what happened. The bot does not attempt
   to merge this PR again until the PR branch or PR target branch change.
   When the bot notices that change, it removes this label.
+* `M-cleared-for-merge`: A human allowed the bot running in
+  `config::staged_run` mode to merge the PR. The label has no effect
+  unless the bot is running in that mode. The bot never sets this label.
+  The bot removes this label after successfully merging the PR.
```
eduard-bagdasaryan: This looks correct to me, but somebody else could have questions: 
* What are differences between 'staged_run' and the label applied and their absence?
* Does the bot only merge the labeled PR (as described) or also other, cleanup steps, completing the PR?

To avoid them, I would rephrase this, emphasizing the fact that the label effectively cancels `staged_run` option (see the similar comment below).

README.md
rousskov: I agree that this could be simplified along the lines you are suggesting. Ignoring the option is slightly imprecise because it is still honored when checking for conflicts among `*_run` options, but I know what you mean.

This change request is no longer applicable because the current PR no longer modifies staged_run functionality, but it was a good comment, thank you.

README.md
rousskov: I could not quite grok your first bullet, but I agree with the second one -- I should not have used the overloaded "merge" term in this context. The current version avoids this problem.

Indirectly, this comment made me realize that the initial approach itself was wrong, thank you. Branch commit 378109e addresses that problem.

rousskov:
FYI: The two TODOs added in branch commit 378109e are to be addressed in other/future PRs, not this one.

eduard-bagdasaryan:
> If updateReference() can detect target branch updates, then please fix this by relying on updateReference() to assert both preconditions.

Pushed 7a49908, following this suggestion

PR #2, opened 2018-02-26
Moved bunyan 'stdout' configuration into config.json

eduard-bagdasaryan:
Added a workaround to support JSON-configurable stdout/stderr streams.

See also: https://github.com/trentm/node-bunyan/issues/153


eduard-bagdasaryan:
Yes, that change was done but missed during 'git add'.

PR #4, opened 2018-02-27
Fixed (end of) line detection in PR descriptions

eduard-bagdasaryan:
GitHub does not normalize line endings in PR descriptions. Its API may
yield CRLF or LF (at least) line endings in the descriptions, probably
depending on how the description was created or edited.

### DIFF
--- a/MergeContext.js
+++ b/MergeContext.js
@@ -513,7 +513,7 @@ class MergeContext {
     }
 
     _prMessageValid() {
-        const lines = this._prMessage().split('\n');
+        const lines = this._prMessage().split(/\r*\n/);
         for (let line of lines) {
             if (line.length > 72)
                 return false;


eduard-bagdasaryan:
> What injects those \r characters on the way from GitHub HTTP response to our code?

In my understanding, these '\r' characters can be simply added by a PR user while editing PR description (using an editor in Windows, e.g.). I don't think that something 'injects' this '\r' while we request PR info with the GitHub api all.

rousskov:
> these '\r' characters can be simply added by a PR user while editing PR description

I suspect you are right. https://api.github.com/repos/squid-cache/squid/pulls/160 is one such example: The current description text there is mine, and I do not use Windows, but it contains `\r\n`, possibly because the newline/encoding was set by the PR _author_. A PR created by me uses `\r\n`: https://api.github.com/repos/squid-cache/squid/pulls/161. This feels like a GitHub web UI bug, but there is nothing we can do about it.

Please remind me: Did you already check that generated PR descriptions containing `\r\n` are committed without `\r`s?

rousskov:
This PR is another example that has `\r\n` newlines even though I have pasted text without them (I could check because I still had that text in my clipboard). Did you create this PR on Windows? If not, then I am guessing that all Web-created PRs have CRs and (some) hub-created PRs do not.

MergeContext.js
```diff
@@ -513,7 +513,7 @@ class MergeContext {
     }
 
     _prMessageValid() {
-        const lines = this._prMessage().split('\n');
+        const lines = this._prMessage().split(/\r*\n/);
```
rousskov: LGTM.

For the record, I tried to handle more encoding cases using a `$` regex specifically designed for end-of-line detection, but `$` does not match the preceding newline sequence itself (e.g., CR or CRLF), matches after CR in CRLF, and overall creates more problems (in this context) than it solves:

```
js> console.log("foo\r\nbar\rbaz\nbog".split(/$/m));
[ 'foo', '\r', '\nbar', '\rbaz', '\nbog' ]
```

eduard-bagdasaryan:
> Please remind me: Did you already check that generated PR descriptions containing \r\n are committed without \rs?

The PR bodies were unchanged. That is fixed now in [PR6](https://github.com/measurement-factory/anubis/pull/6)

eduard-bagdasaryan:
>  Did you create this PR on Windows? If not, then I am guessing that all Web-created PRs have CRs and (some) hub-created PRs do not.

I created PR message in two ways:
* Creating in an editor message with CRLF and copy-pasting into PR web page
* Filling PR message directly on its web page

In both cases, the API call returned CRLF, so your guessing was confirmed.

rousskov:
> The PR bodies were unchanged. That is fixed now in PR #6.

Glad I asked! Thank you for fixing it.

> your guessing was confirmed

or at least semi-confirmed. We have learned something. No more outstanding questions for this PR.

PR #5, opened 2018-02-27
Moved sources into src/

eduard-bagdasaryan:


README.md
```diff
@@ -223,7 +223,9 @@ Properly counting such votes is a missing feature.
 The bot may be started like any node.js script. For example:
 
 ```
-node Main.js /etc/anubis/config.json
+git clone git@github.com:measurement-factory/anubis.git
+cd ./anubis
```
rousskov: I do not think this recipe works well because it skips installation of dependencies (e.g., `npm update`). It is rather unlikely that somebody will have _all_ the right dependencies installed globally. However, I doubt we should tell people how to install Javascript-based software -- there are just too many environment-specific ways to do that.

I suggest fixing the invocation line (like you did) but not giving any other details like `git clone`. The invocation line itself _is_ useful IMO because it is not obvious which `.js` file the user should run or how to specify the configuration file.

eduard-bagdasaryan:
@rousskov, to avoid useless conflicts resolving, I would suggest moving this PR ahead after PR1 is merged. I would create a fresh branch for convenience (which should be merged into master ASAP, to avoid future conflicts).

rousskov:
Agreed, but no need for a new branch. You can force-push into the current PR branch IMO.

rousskov:
This is now the only PR left.

eduard-bagdasaryan:
> You can force-push into the current PR branch

Done.

PR #7, opened 2018-03-21
Concatenate paginated items from Node-Github API calls

eduard-bagdasaryan:
This simple implementation should meet our needs for now.  A smarter
implementation could iterate through paginated items on the fly (not
storing into the memory). However, despite saving some memory
this would complicate the code too much.

### DIFF
--- a/src/GitHubUtil.js
+++ b/src/GitHubUtil.js
@@ -13,16 +13,50 @@ const logApiResult = Log.logApiResult;
 
 const GitHubAuthentication = { type: 'token', username: Config.githubUserLogin(), token: Config.githubToken() };
 
+function commonAppender(allPages, aPage) {
+    allPages.data = allPages.data.concat(aPage.data);
+}
+
+function statusAppender(statusPages, aPage) {
+    let statuses = statusPages.data.statuses;
+    statusPages.data.statuses = statuses.concat(aPage.data.statuses);
+}
+
+function protectedBranchAppender(protectedBranchPages, aPage) {
+    let contexts = protectedBranchPages.data.protection.required_status_checks.contexts;
+    protectedBranchPages.data.protection.required_status_checks.contexts =
+        contexts.concat(aPage.data.protection.required_status_checks.contexts);
+}
+
+async function pager(firstPage, appender) {
+    let allPages = null;
+    if (appender === undefined)
+        appender = commonAppender;
+
+    function doPager (nPage) {
+        if (allPages === null)
+            allPages = nPage;
+        else
+            appender(allPages, nPage);
+
+       if (GitHub.hasNextPage(nPage)) {
+            return GitHub.getNextPage(nPage).then(doPager);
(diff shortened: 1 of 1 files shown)


src/GitHubUtil.js
```diff
@@ -13,16 +13,50 @@ const logApiResult = Log.logApiResult;
 
 const GitHubAuthentication = { type: 'token', username: Config.githubUserLogin(), token: Config.githubToken() };
 
+function commonAppender(allPages, aPage) {
```
rousskov: s/common/default/ so that the reader does not misinterpret "common" as "shared by all".

src/GitHubUtil.js
```diff
@@ -13,16 +13,50 @@ const logApiResult = Log.logApiResult;
...
+}
+
+function statusAppender(statusPages, aPage) {
+    let statuses = statusPages.data.statuses;
+    statusPages.data.statuses = statuses.concat(aPage.data.statuses);
+}
+
+function protectedBranchAppender(protectedBranchPages, aPage) {
+    let contexts = protectedBranchPages.data.protection.required_status_checks.contexts;
+    protectedBranchPages.data.protection.required_status_checks.contexts =
```
rousskov: Why do we need these custom appending functions? Are we doing something wrong elsewhere that makes commonAppender inapplicable to branches and status checks? The official Pagination [example](https://www.npmjs.com/package/@octokit/rest#pagination) does not mention the need for custom appenders but perhaps it is oversimplified.

src/GitHubUtil.js
```diff
@@ -13,16 +13,50 @@ const logApiResult = Log.logApiResult;
...
+        contexts.concat(aPage.data.protection.required_status_checks.contexts);
+}
+
+async function pager(firstPage, appender) {
+    let allPages = null;
+    if (appender === undefined)
+        appender = commonAppender;
+
+    function doPager (nPage) {
```
rousskov: If you can convert this complex-looking recursion into a simple-looking loop as in the Pagination [example](https://www.npmjs.com/package/@octokit/rest#pagination), please do so.

src/GitHubUtil.js
eduard-bagdasaryan: This is needed because the returned by an API call data layout may be different. In the example, each res.data is an array which is concatenated, but statuses and protected branch APIs have paginated arrays somewhere deep inside res.data. Only the caller (i.e., an API context) know what array is going to be concatenated.

src/GitHubUtil.js
eduard-bagdasaryan: I was thinking about that but, unfortunately, have found nothing simpler. We are facing this  'complication' due to a different paginated item layout, mentioned above.

src/GitHubUtil.js
eduard-bagdasaryan: Done.

PR #12, opened 2018-04-12
Treat diverged branches specially when merging staged commit

eduard-bagdasaryan:
When base fast-forwarding fails, a diverged target branch is the only
case where recreating the staged commit is a good idea. Other errors
(e.g., too strict protected branch settings or lack of permissions) are
not good reasons for re-staging: the bot should behave in this case
like in any other PR-specific failure case.

Also print full info for all caught exceptions, including when a
missing. These details could be useful for triage.

### DIFF
--- a/src/Logger.js
+++ b/src/Logger.js
@@ -8,7 +8,7 @@ function LogError(err, context) {
     Log(err, context, "error");
 }
 
-function LogExceptionMessage(err, context) {
+function LogException(err, context) {
     Log(err, context, "info");
 }
 
@@ -31,7 +31,7 @@ function logApiResult(method, params, result) {
 module.exports = {
     Logger: Logger,
     LogError: LogError,
-    LogExceptionMessage: LogExceptionMessage,
+    LogException: LogException,
     logApiResult: logApiResult
 };
 
--- a/src/Logger.js
+++ b/src/Logger.js
@@ -5,12 +5,23 @@ const Config = require('./Config.js');
 const Logger = bunyan.createLogger(Config.loggerParams());
 
 function LogError(err, context) {
+    Log(err, context, "error");
+}
+
+function LogExceptionMessage(err, context) {
+    Log(err, context, "info");
+}
+
+function Log(err, context, kind) {
     assert(context);
     let msg = (err.message === undefined) ? JSON.stringify(err) : err.message;
     msg = context + ": " + msg;
     if (err.stack !== undefined)
         msg += " " + err.stack.toString();
-    Logger.error(msg);
+    if (kind === "error")
+        Logger.error(msg);
+    else
+        Logger.info(msg);
 }
 
 function logApiResult(method, params, result) {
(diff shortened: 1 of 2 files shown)


src/MergeContext.js
```diff
@@ -138,6 +138,17 @@ class MergeContext {
         return tagCommit.tree.sha === prCommit.tree.sha;
     }
 
+    // whether the tag commit and the base HEAD are 'diverged'
```
rousskov: If possible, avoid describing foo() as "does foo". For the reader who does not know what "diverged" means in this context, the current description is not very useful.

s/tag commit/tagged commit/ or s/tag commit/tag/ or /tag commit/staged commit/

src/MergeContext.js
```diff
@@ -351,7 +364,7 @@ class MergeContext {
             requiredContexts = await GH.getProtectedBranchRequiredStatusChecks(this._prBaseBranch());
         } catch (e) {
            if (e.name === 'ErrorContext' && e.notFound())
-               this._log("required status checks not found");
+               Log.LogExceptionMessage(e, this._toString() + " required status checks not found");
```
rousskov: AFAICT, what we want to say here is "no status checks are required". In other words, this is not an error (unlike the implication of "X not found"), and that is why we do not treat it as an error.

src/Logger.js
```diff
@@ -5,12 +5,23 @@ const Config = require('./Config.js');
 const Logger = bunyan.createLogger(Config.loggerParams());
 
 function LogError(err, context) {
+    Log(err, context, "error");
+}
+
+function LogExceptionMessage(err, context) {
```
rousskov: I would call this LogException(), especially since its sister function is _not_ called LogErrorMessage().

rousskov:
> Other errors are not good reasons for tag removing: the bot should behave in this case like for other 'unexpected' errors (i.e., re-try after timeout).

IIRC, PR-specific failures like the "other" FF failure discussed above should not be retried after a timeout. Such errors should lead to the PR being ignored (and, if needed, marked as failed) instead. However, fixing this bug (if a fix is necessary) is outside this issue scope -- the issue _changes_ look correct to me (but the above part of their description does not match my understanding and [documentation](https://github.com/measurement-factory/anubis#error-handling) of how the code should work when things fail).

> Also print full info for caught exceptions, even if this are known as non-error cases, like label missing.

AFAICT, a missing label is never a non-error. It is an error (i.e., an unexpected, negative outcome or condition) but the one we can ignore because we wanted to delete the label in the first place.

I have adjusted the PR description to match my understanding. Please adjust further if needed.

eduard-bagdasaryan:
> Such errors should lead to the PR being ignored (and, if needed, marked as failed) instead.

I think errors like insufficient permissions are general ones and PR-unrelated. Probably the bot should test the account permissions at startup, if possible (and exit if there are such problems) to avoid useless noise in PRs history.

> AFAICT, a missing label is never a non-error. It is an error

Probably my example was not good, because now we check for the label existence before adding/removing. Another example: try to load staging tag for a PR: getReference() throws if there is no such reference.

src/MergeContext.js

src/MergeContext.js

src/Logger.js
eduard-bagdasaryan: Done.

PR #14, opened 2018-04-26
PR approval status checks

eduard-bagdasaryan:
With this change, the bot adds a new 'PR approval' status just after
vote counting. This new status has 'PR approval' name and can have
one of the descriptions:

- waiting for more votes or a slow burner timeout
- waiting for more votes
- waiting for fast track objections
- waiting for requested reviews
- blocked (see change requests)
- approved
- approved (on slow burner)

Each 'waiting', 'blocked' and 'approved' correspond to 'yellow', 'red'
and 'approved' states respectively. This status is added to both PR
and staged commit.  On GitHub, this 'PR approval' check should be
configured as a 'required check'.

By means of this new check, we are free now to keep unchecked
"Require pull request reviews before merging" GitHub protected
branch setting, whereas still requiring (indirectly, via statuses)
PR reviews.

Also: refactored and unified overall check result generation/processing
in MergeContext and PrMerger classes. The result of each step
(startProcessing() and finishProcessing()) has now four independent
states: 'succeeded', 'failed', 'suspended' and 'delayed'. Before this
improvement, many methods returned boolean and additional information
was requested by helper methods, such as MergeContext::delay().

Also fixed 'finishProcessing()' caller context which did not initiate
'rerun' on timeout if checking for approval postconditions require so.

### DIFF
--- a/config-example.json
+++ b/config-example.json
@@ -15,6 +15,7 @@
     "voting_delay_min": "2d",
     "voting_delay_max": "10d",
     "staging_checks": 2,
+    "approval_url": "http://example.com",
     "logger_params" : {
         "name" : "anubis",
         "streams" : [
--- a/config-example.json
+++ b/config-example.json
@@ -15,7 +15,7 @@
     "voting_delay_min": "2d",
     "voting_delay_max": "10d",
     "staging_checks": 2,
-    "approval_url": "http://example.com",
+    "approval_url": "",
     "logger_params" : {
         "name" : "anubis",
         "streams" : [
--- a/src/Config.js
+++ b/src/Config.js
@@ -95,6 +95,15 @@ class ConfigOptions {
     failedDescriptionLabel() { return "M-failed-description"; }
     // allows target branch update in 'guarded_run' mode
     clearedForMergeLabel() { return "M-cleared-for-merge"; }
+
+    approvalUrl() { return "https://github.com/measurement-factory/anubis#voting-and-pr-approvals"; }
+
+    // whether the bot will create 'fake' approval statuses for PR and staged commit.
+    // TODO: make configurable or remove
+    manageApprovalStatus() { return true; }
+
+    // the 'context name' of fake approval status
+    approvalContext() { return "PR approval"; }
 }
 
 const configFile = process.argv.length > 2 ? process.argv[2] : './config.json';
--- a/src/Config.js
+++ b/src/Config.js
@@ -31,6 +31,7 @@ class ConfigOptions {
         this._votingDelayMin = timestring(conf.voting_delay_min, 'ms');
         this._stagingChecks = conf.staging_checks;
         this._loggerParams = conf.logger_params;
+        this._approvalUrl = conf.approval_url;
 
         // unused
         this._githubUserNoreplyEmail = null;
@@ -96,13 +97,13 @@ class ConfigOptions {
     // allows target branch update in 'guarded_run' mode
     clearedForMergeLabel() { return "M-cleared-for-merge"; }
 
-    approvalUrl() { return "https://github.com/measurement-factory/anubis#voting-and-pr-approvals"; }
+    // an URL of the description of the approval test status
+    approvalUrl() { return this._approvalUrl; }
 
-    // whether the bot will create 'fake' approval statuses for PR and staged commit.
-    // TODO: make configurable or remove
-    manageApprovalStatus() { return true; }
+    // whether the bot will create the approval test statuses for PR and staged commit
+    manageApprovalStatus() { return this._approvalUrl.length > 0; }
 
-    // the 'context name' of fake approval status
+    // the 'context name' of the approval test status
     approvalContext() { return "PR approval"; }
 }
 
--- a/src/Config.js
+++ b/src/Config.js
@@ -101,7 +101,7 @@ class ConfigOptions {
     approvalUrl() { return this._approvalUrl; }
 
     // whether the bot will create the approval test statuses for PR and staged commit
-    manageApprovalStatus() { return this._approvalUrl.length > 0; }
+    manageApprovalStatus() { return this.approvalUrl().length > 0; }
 
     // the 'context name' of the approval test status
     approvalContext() { return "PR approval"; }
--- a/src/GitHubUtil.js
+++ b/src/GitHubUtil.js
@@ -384,7 +384,7 @@ function createStatus(sha, state, targetUrl, desc, context) {
     params.sha = sha;
     params.state = state;
     params.target_url = targetUrl;
-    params.desc = desc;
+    params.description= desc;
     params.context = context;
     return new Promise( (resolve, reject) => {
       GitHub.authenticate(GitHubAuthentication);
--- a/src/GitHubUtil.js
+++ b/src/GitHubUtil.js
@@ -143,6 +143,7 @@ function getStatuses(ref) {
             }
             res = await pager(res, statusAppender);
             logApiResult(getStatuses.name, params, {statuses: res.data.statuses.length});
+            assert(res.data.state === 'success' || res.data.state === 'pending' || res.data.state === 'failure');
             resolve(res.data);
         });
     });
--- a/src/GitHubUtil.js
+++ b/src/GitHubUtil.js
@@ -378,13 +378,13 @@ function removeLabel(label, prNum) {
 //    });
 //}
 
-function createStatus(sha, state, targetUrl, desc, context) {
+function createStatus(sha, state, targetUrl, description, context) {
     assert(!Config.dryRun());
     let params = commonParams();
     params.sha = sha;
     params.state = state;
     params.target_url = targetUrl;
-    params.description= desc;
+    params.description = description;
     params.context = context;
     return new Promise( (resolve, reject) => {
       GitHub.authenticate(GitHubAuthentication);
--- a/src/MergeContext.js
+++ b/src/MergeContext.js
@@ -9,52 +9,50 @@ const Util = require('./Util.js');
 class StepResult
 {
     // treat as private; use static methods below instead
-    constructor(delayMs, suspended) {
+    constructor(delayMs) {
         assert(delayMs !== undefined);
-        assert(suspended !== undefined);
         this._delayMs = delayMs;
-        this._suspended = suspended;
     }
 
     // the step is successfully finished
     static Succeed() {
-        return new StepResult(0, null);
+        return new StepResult(0);
     }
 
     // the step is postponed (and will be resumed some time later)
     static Suspend() {
-        return new StepResult(null, true);
+        return new StepResult(-1);
     }
 
     // the step is finished with failure
     static Fail() {
-        return new StepResult(null, null);
+        return new StepResult(null);
     }
 
     // the step is postponed (and will be resumed in delayMs)
     static Delay(delayMs) {
-        assert(delayMs > 0);
-        return new StepResult(delayMs, null);
+        assert(delayMs !== null && delayMs > 0);
+        return new StepResult(delayMs);
     }
 
     succeeded() {
-        return this._delayMs === 0 && this._suspended === null;
+        return this._delayMs !== null && this._delayMs === 0;
     }
 
     failed() {
-        return this._delayMs === null && this._suspended === null;
+        return this._delayMs === null;
     }
 
     suspended() {
-        return this._delayMs === null && this._suspended === true;
+        return this._delayMs !== null && this._delayMs < 0;
     }
 
     delayed() {
-        return this._delayMs > 0 && this._suspended === null;
+        return this._delayMs !== null && this._delayMs > 0;
     }
 
     delay() {
-        assert(this._delayMs > 0);
+        assert(this.delayed());
         return this._delayMs;
     }
 }
@@ -101,7 +99,12 @@ class Approval {
 
     grantedTimeout() { return this.delayMs > 0; }
 
-    toString() { return "description: " + this.description + ", state: " + this.state; }
+    toString() {
+        let str = "description: " + this.description + ", state: " + this.state;
+        if (this.delayMs !== null)
+            str += ", delayMs: " + this.delayMs;
+        return str;
+    }
 }
 
 // Processing a single PR
@@ -726,7 +729,7 @@ class MergeContext {
     _dryRun(msg) {
         if (!Config.dryRun())
             return false;
-        this._log("skip " + msg + " due to dry_run option");
+        this._log("skip '" + msg + "' due to dry_run option");
         return true;
     }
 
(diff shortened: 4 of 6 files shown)


src/Config.js
```diff
@@ -95,6 +95,15 @@ class ConfigOptions {
...
     // allows target branch update in 'guarded_run' mode
     clearedForMergeLabel() { return "M-cleared-for-merge"; }
+
+    approvalUrl() { return "https://github.com/measurement-factory/anubis#voting-and-pr-approvals"; }
+
+    // whether the bot will create 'fake' approval statuses for PR and staged commit.
+    // TODO: make configurable or remove
+    manageApprovalStatus() { return true; }
+
+    // the 'context name' of fake approval status
```
rousskov: s/fake approval/the approval test/. The test can be considered fake (in some sense), but the status is real. I do not think we should use this vague "fakeness" notion in the bot code.

src/MergeContext.js
```diff
@@ -298,11 +303,13 @@ class MergeContext {
         const collaborators = await GH.getCollaborators();
         const pushCollaborators = collaborators.filter(c => c.permissions.push === true);
         const requestedReviewers = this._prRequestedReviewers();
+        this._approvalResult = {params: null, delayMs: -1};
```
rousskov: I see no reason to introduce a magic numerical constant instead of using a much more descriptive/appropriate `null`. If there are not good reasons for `-1` please rewrite to use `null` instead.

src/MergeContext.js
```diff
@@ -311,7 +318,9 @@ class MergeContext {
         const prAgeMs = new Date() - new Date(this._createdAt());
         if (prAgeMs < Config.votingDelayMin()) {
             this._log("in minimal voting period");
-            return Config.votingDelayMin() - prAgeMs;
+            this._approvalResult.params = {desc: "waiting for initial review", state: "pending"};
```
rousskov: This description may not match reality -- we may already have 10 reviews here. I think this check has to go lower so that we deemphasize the unimportant time aspect when there are not enough votes. In that lower position, the descriptions would become something like `waiting for fast track objections` and `waiting for more votes or a slow burner timeout`, depending on whether we already have sufficient or just necessary votes.

src/MergeContext.js
```diff
@@ -337,19 +346,47 @@ class MergeContext {
...
+        if (prAgeMs >= Config.votingDelayMax()) {
+            this._approvalResult.params = {desc: "approved (on slow burner)", state: "success"};
+            this._approvalResult.delayMs = 0;
+            return;
         }
-        if (usersApproved.length >= Config.sufficientApprovals() || prAgeMs >= Config.votingDelayMax())
-            return 0;
         this._log("in maximum voting period");
-        return Config.votingDelayMax() - prAgeMs;
+        this._approvalResult.params = {desc: "waiting for time (slow burner)", state: "pending"};
```
rousskov: Please use `waiting for more votes or a slow burner timeout` instead.

src/GitHubUtil.js
```diff
@@ -384,7 +384,7 @@ function createStatus(sha, state, targetUrl, desc, context) {
     params.sha = sha;
     params.state = state;
     params.target_url = targetUrl;
-    params.desc = desc;
+    params.description= desc;
```
rousskov: Please rename the formal parameter (`desc`) as well.

src/MergeContext.js
```diff
@@ -14,6 +14,8 @@ class MergeContext {
         this._shaLimit = 6;
         // the remainder (>0) of the min or max voting delay (in ms)
         this._votingDelay = null;
+        // information used for fake approval status creation/updating
+        this._approvalResult = {params: null, delayMs: -1};
```
rousskov: Please create an Approval class (with a proper constructor, etc.) instead of using these nested anonymous structs.

src/MergeContext.js
```diff
@@ -222,7 +226,7 @@ class MergeContext {
             }
         }
 
-        this._votingDelay = delay;
+        this._votingDelay = this._approvalResult.delayMs;
```
rousskov: If `this._votingDelay` is always `this._approvalResult.delayMs`, then we should remove the former or transform it into a method.

src/MergeContext.js
```diff
@@ -337,19 +346,47 @@ class MergeContext {
...
         this._log("approved by " + usersApproved.length + " core developer(s)");
 
         if (usersApproved.length < Config.necessaryApprovals()) {
             this._log("not approved by necessary " + Config.necessaryApprovals() + " votes");
-            return null;
+            this._approvalResult.params = {desc: "waiting for more votes", state: "pending"};
+            return;
+        }
+        if (usersApproved.length >= Config.sufficientApprovals()) {
+            this._approvalResult.params = {desc: "approved", state: "success"};
```
rousskov: It is strange that approved PRs do not have a `_log()` call while other `if`s have it. However, it may be better to simply log the `_approvalResult` upon this method return (in all cases).

src/MergeContext.js
```diff
@@ -298,11 +303,13 @@ class MergeContext {
         const collaborators = await GH.getCollaborators();
         const pushCollaborators = collaborators.filter(c => c.permissions.push === true);
         const requestedReviewers = this._prRequestedReviewers();
+        this._approvalResult = {params: null, delayMs: -1};
 
         for (let collaborator of pushCollaborators) {
             if (requestedReviewers.includes(collaborator.login)) {
                 this._log("requested core reviewer: " + collaborator.login);
-                return null;
+                this._approvalResult.params = {desc: "blocked (see review requests)", state: "error"};
```
rousskov: Please use `waiting for requested reviews` description instead.

rousskov: Like other "waiting" statuses, this should be a `pending`, not `error` status AFAICT (even if GitHub treats review requests as a "red" problem). The `error` status should be used only when there is something wrong with the pull request.

config-example.json
```diff
@@ -15,6 +15,7 @@
     "voting_delay_min": "2d",
     "voting_delay_max": "10d",
     "staging_checks": 2,
+    "approval_url": "http://example.com",
```
rousskov: Please use an empty string (disabling the feature) rather than wrong URL by default. Please do not forget to update README as well.

src/Config.js
```diff
@@ -95,6 +96,15 @@ class ConfigOptions {
     failedDescriptionLabel() { return "M-failed-description"; }
     // allows target branch update in 'guarded_run' mode
     clearedForMergeLabel() { return "M-cleared-for-merge"; }
+
+    // an URL of the description of the approval test status
+    approvalUrl() { return this._approvalUrl; }
+
+    // whether the bot will create the approval test statuses for PR and staged commit
+    manageApprovalStatus() { return this._approvalUrl.length > 0; }
```
rousskov: s/this._approvalUrl/approvalUrl()/ to minimize low-level member usage.

src/MergeContext.js
```diff
@@ -4,6 +4,26 @@ const Log = require('./Logger.js');
...
+    }
+
+    reset() {
+        this.assign(null, null, null);
+    }
+
+    assign(aDesc, aState, aDelayMs) {
+        this.description = aDesc;
+        this.state = aState;
+        // If approved(slow burner): the number for milliseconds to wait for
```
rousskov: We should not consider PRs waiting on slow burner approved. Also, PRs on slow burner should be treated very similar to PRs on the fast track that are waiting for their timeout. but the proposed comment items lack that symmetry.

How about the following instead?

```
// If waiting for a timeout to merge the staged commit: > 0
// If ready to merge the staged commit now: == 0
// Otherwise (e.g., failed description, negative votes, or review requested): null
```

src/MergeContext.js
```diff
@@ -198,6 +218,11 @@ class MergeContext {
             return false;
         }
 
+        await this._checkApproved();
+        this._log("_checkApproved: " + this._approvalResult.description + ", state: " +
```
rousskov: Please move this "serialization" into Approval::toString() so that it is called automatically and can easily be more flexible (e.g., does not have to print nil delays).

src/MergeContext.js
```diff
@@ -209,8 +234,7 @@ class MergeContext {
             return false;
         }
 
-        const delay = await this._checkApproved();
-        if (delay === null) {
+        if (this._approvalResult.delayMs === null) {
```
rousskov: Please convert that logic into a method. For example, Approval::blocked().

src/MergeContext.js
```diff
@@ -292,28 +315,23 @@ class MergeContext {
...
         const collaborators = await GH.getCollaborators();
         const pushCollaborators = collaborators.filter(c => c.permissions.push === true);
         const requestedReviewers = this._prRequestedReviewers();
+        this._approvalResult.reset();
 
         for (let collaborator of pushCollaborators) {
             if (requestedReviewers.includes(collaborator.login)) {
                 this._log("requested core reviewer: " + collaborator.login);
-                return null;
+                this._approvalResult.assign("waiting for requested reviews", "pending", null);
```
rousskov: I think we should split assign() into two or three methods so that the code intent is clear and so that the second method can check for a non-null delay (while the other two do not need to bother with a delay). For example, Approval::block(), Approval::grantAfterTimeout(), and Approval::grantNow().

BTW, I am not yet sure that zero timeout needs special handling -- I am just trying to go with what you have proposed.

src/MergeContext.js
```diff
@@ -337,19 +355,55 @@ class MergeContext {
...
+            return;
+        if (!Config.manageApprovalStatus())
+            return;
+
+        assert(this._approvalResult);
+        let combinedStatus = await GH.getStatuses(sha);
+        const approvalStatus = combinedStatus.statuses ?
+            combinedStatus.statuses.find(el => el.context.trim() === Config.approvalContext()) : null;
+
+        if (approvalStatus && approvalStatus.state === this._approvalResult.state &&
```
rousskov: I would move the two comparisons into Approval::matchesGitHubStatusCheck() or similar. As a bonus, that method can check that `this.state` and `this.description` are not null.

src/MergeContext.js
```diff
@@ -12,8 +32,8 @@ class MergeContext {
         this._pr = pr;
         this._tagSha = (tSha === undefined) ? null : tSha;
         this._shaLimit = 6;
-        // the remainder (>0) of the min or max voting delay (in ms)
-        this._votingDelay = null;
+        // information used for approval test status creation/updating
+        this._approvalResult = new Approval();
```
rousskov: Please consider s/_approvalResult/_approval/.

eduard-bagdasaryan:
While addressing review comments, I found and fixed two additional problems, described in second 'also' of d572ca0 and in d572ca0 preambles.

src/MergeContext.js
```diff
@@ -198,6 +240,12 @@ class MergeContext {
             return false;
         }
 
+        const delay = await this._checkApproved();
```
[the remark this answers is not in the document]
rousskov: I would also rename `_checkApproved()` to `_checkApproval()` or `_computeApproval()` but that is a bit out of scope here. Your call.

src/MergeContext.js
```diff
@@ -409,13 +485,12 @@ class MergeContext {
                 if (requiredChecks.find(el => el.context.trim() === requiredContext.trim()) === undefined) {
                     // but the passed check matches a required one
```
rousskov: Please note that the following combination of comments does not make sense:

```
// a passed check misses a required one
// but the passed check matches a required one
```
I cannot suggest how to fix this because I am not sure what this all means. Some kind of `context` vs `check` typo?

src/MergeContext.js
```diff
@@ -4,6 +4,42 @@ const Log = require('./Logger.js');
...
+    }
+
+    reset() {
+        this.description = null;
+        this.state = null;
+        this._blocked = true;
+    }
+
+    allow(aDesc) {
+        this.description = aDesc;
```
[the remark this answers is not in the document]
eduard-bagdasaryan: I added assertions for the description and state into the constructor.

src/MergeContext.js
```diff
@@ -405,17 +474,16 @@ class MergeContext {
         const ret = (prevLen === requiredChecks.length);
         if (ret && andSupplyRequired) {
             for (let requiredContext of requiredContexts) {
-                // a passed check misses a required one
+                // go further only if the passed check context is not a required one
                 if (requiredChecks.find(el => el.context.trim() === requiredContext.trim()) === undefined) {
-                    // but the passed check matches a required one
+                    // go further only if the passed check context matches the required one
```
```
// go further only if the passed check context is not a required one 
// go further only if the passed check context matches the required one 
```

It looks like we are now implying that a "passed check context" may "not be a required one" but may "match the required one", which is still very confusing because we should not even bother with non-required checks. The whole intent of this code is unclear and the comments are not helping. Perhaps there is a wider terminology/naming problem. Perhaps the overall intent should be documented _before_ the loop, using consistent terminology. Unfortunately, I do not understand this code enough to suggest a specific fix.

I cannot spend more time on this mess. You can ignore this complaint if you are sure the code "works".

src/MergeContext.js
```diff
@@ -548,7 +616,7 @@ class MergeContext {
 
     // the processing of this PR is delayed on this
     // number of milliseconds
-    delay() { return this._votingDelay; }
+    delay() { return this._approval === null ? null : this._approval.delayMs; }
```
rousskov: This looks unsafe to me. If possible, I would like to see something closer to this:

```
delay() {
     // without computing the approval status, we cannot know about delays
    assert(this._approval);
    return this._approval.delay();
}
```

where Approval::delay() asserts that either the approval was granted (now or after a timeout) OR the approval was not blocked/suspended.

The overall intent here remains the same: Make sure that Approval states are mutually exclusive so that we cannot accidentally merge a blocked or suspended PR just because our `!delay()` check succeeded (with a nil delay).

src/MergeContext.js
```diff
@@ -4,6 +4,43 @@ const Log = require('./Logger.js');
...
+    static Block(description) {
+        return new Approval(description, "pending", null);
+    }
+
+    matchesGitHubStatusCheck(approvalStatus) {
+        assert(approvalStatus);
+        return approvalStatus.state === this.state && approvalStatus.description === this.description;
+    }
+
+    blocked() { return this.delayMs === null; }
```
rousskov: With the other changes discussed elsewhere applied, this may become something like return this.state === "failure" (or whatever the "red" state name is).

* blocked(): Cannot be merged unless GitHub state changes. No delay. Red status -- somebody needs to fix something.
* suspended(): Cannot be merged unless GitHub state changes. No delay. Yellow status -- somebody needs to add something.
* otherwise: Can be merged:
  * either immediately (i.e., zero delay()) (green status) or
  * after a positive delay() (yellow status).

I do not know whether the high-level Approval users need to know the difference between blocked() and suspended(). If they do not, then the two methods can be replaced with a single `!granted()` check. The Approval-creating code does know the difference, of course, so there will be two different static methods: Block() and Suspend().

PR #15, opened 2018-05-04
Always refresh approval statuses and M-failed-description label

eduard-bagdasaryan:
... except cases when the PR is already closed or merged.  Before this
fix, checkMergeConditions() exited immediately when it stumbled upon a
failed condition. Obviously, before exiting, we need to sync independent
PR attributes (such as M-failed-description label or approval statuses),
since we request this information anyway.

Also: undone 0d00635 fix (the second 'also'). At finishProcessing()
stage, we cannot wait for a delay (which could appear, if the bot
time-voting settings have been changed), because these would
block other PRs. So, in such situation, we have to cleanup and let others
proceed.

Also: a couple of out-of-scope polishing fixes.

### DIFF
--- a/src/MergeContext.js
+++ b/src/MergeContext.js
@@ -74,6 +74,7 @@ class Approval {
     }
 
     static GrantAfterTimeout(description, delayMs) {
+        assert(delayMs);
         assert(delayMs > 0);
         return new Approval(description, "pending", delayMs);
     }
@@ -97,7 +98,7 @@ class Approval {
 
     granted() { return this.delayMs !== null; }
 
-    grantedTimeout() { return this.delayMs > 0; }
+    grantedTimeout() { return this.delayMs !== null && this.delayMs > 0; }
 
     toString() {
         let str = "description: " + this.description + ", state: " + this.state;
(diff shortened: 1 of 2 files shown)


src/MergeContext.js
```diff
@@ -318,13 +301,35 @@ class MergeContext {
...
+            if (await this._stagingFailed()) {
+                this._log(what + " 'fresh tag with failed staging checks' failed'");
+                return StepResult.Fail();
+            }
+        }
+
         const commitStatus = await this._checkStatuses(this._prHeadSha());
-        /// status checks either failed, or pending and we don't know the time to rerun in
+        /// status checks either failed, or pending (except the approval check, having timeout)
         if (commitStatus === 'failure' || (commitStatus === 'pending' && !this._approval.grantedTimeout())) {
```
rousskov: I am not sure (because I cannot fully understand _checkStatuses() details), but it seems possible that commitStatus is `pending` because there is a real pending test _and_ a pending approval? In that case, the second condition will fail because this._approval.grantedTimeout() will be true.

If there is a bug here, then fixing it may be outside this PR scope. Your call.

eduard-bagdasaryan: > but it seems possible that commitStatus is pending because there is a real pending test and a pending approval

Yes, it is possible we have 'pending approval' (e.g., 'fast track', grantedTimeout() is true), and some of the checks are 'pending'. This case will be processed similarly as if no other 'pending' checks exist: the PR processing will finish, enabling rerun in 'fast track' time left. I see no bug here.

rousskov: I will simplify the presentation a little by ignoring message validity and this._approval.granted() checks after the complex `if` statement.

To me, the code in question reads like this:

```
if (some test has failed)
    return failure;

if (some test is pending and it is not our fake test that waits for objections)
    return failure;

/* all real tests are successful, merge now or wait more for objections */
...
```

which would be OK. However, AFAICT, you are saying that the code is actually doing this:

```
if (some test failed)
    return failure;

if (some tests are pending but we are not waiting for objections)
    return failure;

/* XXX: some real tests may be still running here, but we ignore them */
/* merge now or wait for objections */
...
```

I do not know whether the XXX above leads to what you might call "real bugs", but it feels dangerously wrong at best.

eduard-bagdasaryan: I tried to address your concerns in [PR17](https://github.com/measurement-factory/anubis/pull/17/files).

PR #16, opened 2018-05-04
Process cleared for merge PRs first; fixed timeouts

eduard-bagdasaryan:
When possible, select the smallest PR cleared for merging instead of the
smallest PR that is otherwise ready for merge but has not been cleared
by a human yet. This is what users naturally want.

Side effect: Cleared PRs will have a priority in unguarded run mode.

XXX: guarded run mode: if there are no cleared PRs, we still start
working on an otherwise ready PR, often creating unwanted delays.
TODO: Abandon wrong guesses.

Also fixed rerun timeout calculation: This timeout should be calculated
as the minimum delay among delayed PRs.

### DIFF
--- a/src/PrMerger.js
+++ b/src/PrMerger.js
@@ -17,6 +17,27 @@ class PrMerger {
         this.rerunIn = null;
     }
 
+    async _clearedForMerge(prNum) {
+        const labels = await GH.getLabels(prNum);
+        return labels.find(lbl => lbl.name === Config.clearedForMergeLabel()) !== undefined;
+    }
+
+    async _getPRList() {
+        let prList = await GH.getPRList();
+        for (let pr of prList)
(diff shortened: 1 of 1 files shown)


rousskov:
@eduard-bagdasaryan, please see whether the following commit message is more accurate/useful:

----

Process cleared for merge PRs first and fixed timeouts

When possible, select the smallest PR cleared for merging instead of the
smallest PR that is otherwise ready for merge but has not been cleared
by a human yet. This is what users naturally want.

Side effect: Cleared PRs will have a priority in unguarded run mode.

XXX: If there are no cleared PRs, we still start working on an otherwise
ready PR, often creating unwanted delays. TODO: Abandon wrong guesses.

Also fixed rerun timeout calculation: This timeout should be calculated
as the minimum delay among delayed PRs.

PR #17, opened 2018-05-05
Improved and fixed status checks processing

eduard-bagdasaryan:
Added a new StatusResult class having all required methods for status
checks analysis. With it, fixed and clarified status checking algorithm.
Before this fix, the bot could ignore required checks after changing
GitHub configuration  (or after adding new staged checks). To make
statuses checking more clear, refactored, separating the code into
two cases:

1. PR-statuses (_checkPRStatuses()). The list of checks is verified
against GtiHub-required checks list.

2. Staging-statuses  (_processStagingStatuses()). All staging checks are
assumed 'required'. The bot needs at least 'staging_checks' number of
checks for the staging commit. When all staging checks succeed, one
more step is performed: applying GitHub-required checks to the staged
commit. This step is needed to overcome GitHub protection for staged
commits.

Also: fixed approval status calculation within fast track period. During
that period, if vote number is less than sufficient_approvals, the status
should be 'waiting for more votes or a slow burner timeout', instead of
'waiting for fast track objections'.

Also: copy missing PR statuses into staging commit (instead of coping
'matching' staging statuses). Disadvantage: we mark staging commits
with unrelated (i.e. related to PR branch only) checks. Advantage: this
will work when we have different checks for PR and staging.

### DIFF
--- a/src/MergeContext.js
+++ b/src/MergeContext.js
@@ -110,68 +110,37 @@ class Approval {
 
 class StatusCheck
 {
-    constructor(context, state, targetUrl, description, required) {
+    constructor(context, state, targetUrl, description) {
         assert(context);
         assert(state);
         assert(targetUrl);
         assert(description);
-        assert(required !== undefined && required !== null);
 
         this.context = context;
         this.state = state;
         this.targetUrl = targetUrl;
         this.description = description;
-        this.required = required;
     }
 }
 
 // Passed status checks analysis
 class StatusResult
 {
-    // checksNumber: for staged commits: the bot-configured number of checks (Config.stagingChecks()),
-    //   for PR commits: GitHub configured number of required checks
-    // requiredContexts: GitHub configured number of required checks
-    constructor(checksNumber, requiredContexts) {
+    // checksNumber:
+    //   for staged commits: the bot-configured number of required checks (Config.stagingChecks()),
+    //   for PR commits: GitHub configured number of required checks (requested from GitHub)
+    // context: either "PR" or "Staging";
+    constructor(checksNumber, context) {
         assert(checksNumber);
-        assert(requiredContexts);
-        this.statusChecks = [];
+        assert(context);
         this.checksNumber = checksNumber;
-        this.requiredContexts = requiredContexts;
+        this.context = context;
+        this.statusChecks = [];
     }
 
     addStatus(statusCheck) {
         assert(statusCheck);
         this.statusChecks.push(statusCheck);
-    }
-
-    requiredChecksPassed() {
-        let n = 0;
-        for (let check of this.statusChecks)
-            if (check.required)
-                n++;
-        return n;
-    }
-
-    // whether we have all necessary checks (including all required ones) and some them are pending
-    pending() {
-        assert(this.requiredContexts);
-        return (this.somePending() || this.requiredChecksPassed() < this.requiredContexts);
-    }
-
-    // whether all necessary checks (including all required ones) finished and some of them failed
-    failed() {
-        if (this.pending()) {
-            // some of checks may be failed already, but we need to wait them all finishing
-            return false;
-        }
-        return this.someFailed();
-    }
-
-    // whether all necessary checks (including all required ones) successfully finished
-    succeeded() {
-        if (this.pending())
-            return false;
-        return this.someSucceeded();
     }
 
     // there are some pending contexts except contextName
@@ -189,7 +158,7 @@ class StatusResult
     }
 
     // whether at we have at least checksNumber checks and some of them are pending
-    somePending() {
+    pending() {
         if (this.statusChecks.length < this.checksNumber)
             return true;
         if (this.statusChecks.find(check => check.state === 'pending'))
@@ -197,26 +166,21 @@ class StatusResult
         return false;
     }
 
-    // whether at least checksNumber checks finished and some of them failed
-    someFailed() {
-        if (this.somePending()) {
-            // some of checks may be failed already, but we need to wait them all finishing
-            return false;
-        }
+    // whether some of checks failed
+    failed() {
         return this.statusChecks.find(check => check.state === 'failure' || check.state === 'error') !== undefined;
     }
 
-    // whether at least checksNumber checks finished and all of them succeeded
-    someSucceeded() {
-        if (this.somePending() || this.someFailed())
+    // whether at least checksNumber checks finished and none of them failed
+    succeeded() {
+        if (this.pending() || this.failed())
             return false;
-        // all checks have passed, and none of them is 'pending', 'failure' or 'error'
         return true;
     }
 
     toString() {
-        let combinedStatus = "some/finished: " + this.checksNumber + "/" + this.statusChecks.length +
-            " requiredConfigured/requiredCompleted: " + this.requiredContexts + "/" + this.requiredChecksPassed() + ", combined: '";
+        let combinedStatus = "context: '" + this.context + "' expected/received: " + this.checksNumber + "/" +
+            this.statusChecks.length + ", combined: '";
         if (this.pending())
             combinedStatus += "pending'";
         else if (this.failed())
@@ -299,18 +263,19 @@ class MergeContext {
         // cannot be 'diverged' because _needRestart() succeeded
         assert(compareStatus === "ahead");
 
-        const commitStatus = await this._checkStatuses(this._tagSha, Config.stagingChecks(), true);
+        const commitStatus = await this.__checkStagingStatuses();
         this._log("status details: " + commitStatus);
-        if (commitStatus.pending()) {
-            this._log("waiting for more staging checks completing");
-            return StepResult.Suspend();
-        } else if (commitStatus.failed()) {
+        if (commitStatus.failed()) {
             this._log("staging checks failed");
             return await this._cleanupMergeFailed(false, this._labelFailedStagingChecks);
+        } else if (commitStatus.pending()) {
+            this._log("waiting for more staging checks completing");
+            return StepResult.Suspend();
+        } else {
+            assert(commitStatus.succeeded());
+            await this._supplyStagingWithPrRequired(commitStatus);
+            this._log("staging checks succeeded");
         }
-
-        assert(commitStatus.succeeded());
-        this._log("staging checks succeeded");
 
         if (this._dryRun("finish processing"))
             return StepResult.Suspend();
@@ -349,7 +314,7 @@ class MergeContext {
         const isFresh = await this._tagIsFresh();
         this._log("staging tag is " + (isFresh ? "fresh" : "stale"));
         if (isFresh) {
-            const commitStatus = await this._checkStatuses(this._tagSha, Config.stagingChecks());
+            const commitStatus = await this.__checkStagingStatuses();
             this._log("status details: " + commitStatus);
             if (commitStatus.failed()) {
                 this._log("staging checks failed some time ago");
@@ -446,7 +411,7 @@ class MergeContext {
             }
         }
 
-        const commitStatus = await this._checkStatuses(this._prHeadSha());
+        const commitStatus = await this._checkPRStatuses();
         this._log("status details: " + commitStatus);
         /// status checks either failed, or pending (except the approval check, having timeout)
         if (commitStatus.failed() || commitStatus.othersPending(Config.approvalContext) ||
@@ -612,11 +577,8 @@ class MergeContext {
         await GH.createStatus(sha, this._approval.state, Config.approvalUrl(), this._approval.description, Config.approvalContext());
     }
 
-    // returns filled StatusResult object
-    // checksNumber: the explicit number of requires status checks
-    // andSupplyRequired: if provided, append missing 'required' statuses
-    // to the ref (if the ref has the corresponding matching status already)
-    async _checkStatuses(ref, checksNumber, andSupplyRequired) {
+    // returns filled StatusResult object or throws
+    async _checkPRStatuses() {
         let requiredContexts;
         try {
             requiredContexts = await GH.getProtectedBranchRequiredStatusChecks(this._prBaseBranch());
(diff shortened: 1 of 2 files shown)


src/MergeContext.js
```diff
@@ -173,18 +263,20 @@ class MergeContext {
...
+        if (commitStatus.failed()) {
             this._log("staging checks failed");
             return await this._cleanupMergeFailed(false, this._labelFailedStagingChecks);
+        } else if (commitStatus.pending()) {
+            this._log("waiting for more staging checks completing");
+            return StepResult.Suspend();
+        } else {
+            assert(commitStatus.succeeded());
+            await this._supplyStagingWithPrRequired(commitStatus);
+            this._log("staging checks succeeded");
```
eduard-bagdasaryan: @rousskov, your [algorithm](https://measurementfactory.atlassian.net/browse/SQUID-330?focusedCommentId=39140&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-39140) for staged commits is implemented here: (1) at 268, (2) and (3) at 271, extra step (Copy all required recorded PR checks...) at 276.

src/MergeContext.js
```diff
@@ -499,58 +588,73 @@ class MergeContext {
...
+            if (requiredContexts.find(el => el.trim() === st.context.trim()) !== undefined)
+                statusResult.addStatus(new StatusCheck(st.context, st.state, st.target_url, st.description));
+        }
+        return statusResult;
+    }
+
+    // returns filled StatusResult object
+    async __checkStagingStatuses() {
+        let combinedStatus = await GH.getStatuses(this._tagSha);
+        // assert(combinedStatus.statuses.length <= Config.stagingChecks());
```
eduard-bagdasaryan: Unfortunately, this assert cannot work here (checking that no new staging checks exceeding our 'stagingChecks()' parameter appeared). The reason is that for guarded_run mode we get here again, already after _supplyStagingWithPrRequired() created extra checks (PR checks for staged commit).

src/MergeContext.js
```diff
@@ -108,6 +108,96 @@ class Approval {
...
+    // whether at we have at least checksNumber checks and some of them are pending
+    pending() {
+        if (this.statusChecks.length < this.checksNumber)
+            return true;
+        if (this.statusChecks.find(check => check.state === 'pending'))
+            return true;
+        return false;
+    }
+
+    // whether some of checks failed
```
rousskov: I would be slightly more precise: `at least one check has failed or erred`. However, if possible, we should make this method more reliable/robust by checking that `state` is neither `pending` nor `success`. In that case, the description could say `something went wrong with at least one of the status checks`.

src/MergeContext.js
```diff
@@ -108,6 +108,96 @@ class Approval {
...
+    failed() {
+        return this.statusChecks.find(check => check.state === 'failure' || check.state === 'error') !== undefined;
+    }
+
+    // whether at least checksNumber checks finished and none of them failed
```
rousskov: The description does not match the implementation at all. The implementation is correct.

src/MergeContext.js
```diff
@@ -108,6 +108,96 @@ class Approval {
...
+    succeeded() {
+        if (this.pending() || this.failed())
```
rousskov: After the changes discussed elsewhere in this review, and after making the implementation more reliable/robust, this method would look similar to the following:

```
// the results are final and all checks were a success
succeeded() {
    return this.final() && !this.statusChecks.find(check => check.state !== 'success');
}
```

src/MergeContext.js
```diff
@@ -108,6 +108,96 @@ class Approval {
...
+    // checksNumber:
+    //   for staged commits: the bot-configured number of required checks (Config.stagingChecks()),
+    //   for PR commits: GitHub configured number of required checks (requested from GitHub)
+    // context: either "PR" or "Staging";
+    constructor(checksNumber, context) {
+        assert(checksNumber !== undefined && checksNumber !== null);
+        assert(context);
+        this.checksNumber = checksNumber;
+        this.context = context;
+        this.statusChecks = [];
```
rousskov: Please rename `this.statusChecks` to `this.requiredChecks`. Otherwise, most methods using this data member look really buggy!

src/MergeContext.js
```diff
@@ -108,6 +108,96 @@ class Approval {
...
+    }
+
+    addStatus(statusCheck) {
```
rousskov: Please rename to addRequiredStatus to emphasize that the caller must filter out the tests that may fail.

src/MergeContext.js
```diff
@@ -499,58 +559,115 @@ class MergeContext {
...
+        }
+        return statusResult;
+    }
+
+    async _checkPRStatuses(logContext) {
+        const commitStatus = await this._getPRStatuses();
+        this._log("status details: " + commitStatus);
+        if (commitStatus.succeeded())
+            return StepResult.Succeed();
+        else if (this._approval.grantedTimeout() && commitStatus.approvalPending())
```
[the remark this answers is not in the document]
eduard-bagdasaryan: We do not wait if a test has failed: commitStatus.approvalPending() should return false in this case.

src/MergeContext.js
```diff
@@ -108,6 +108,87 @@ class Approval {
...
+        this.context = context;
+        this.requiredChecks = [];
+    }
+
+    addRequiredStatus(requiredCheck) {
+        assert(requiredCheck);
+        this.requiredChecks.push(requiredCheck);
+    }
+
+    approvalPending() {
```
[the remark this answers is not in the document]
eduard-bagdasaryan: This method is needed to distinguish 2 kinds of pending checks: suspended and delayed. For now, the only delayed check is 'approval' check. For simplicity, I created a dedicated method (named specifically), however, probably there should be a general-purpose method with a parameter (a check name).

src/MergeContext.js
```diff
@@ -499,58 +559,115 @@ class MergeContext {
...
+        if (!requiredContexts)
+            return;
+
+        let requiredPRs = 0;
+        for (let requiredContext of requiredContexts) {
+            if (statusResult.requiredChecks.some(el => el.context.trim() === requiredContext.trim()))
+                requiredPRs++;
+            else {
+                // go further only if the passed check context matches the required one
+                const matched = statusResult.requiredChecks.find(el => el.context.startsWith(requiredContext.trim()));
```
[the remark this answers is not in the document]
eduard-bagdasaryan: We need this 'matched' in order to copy from it different attributes, like url, description, etc. I wonder how can we 'siimply copy' if we do not have information we need for the copy.

src/MergeContext.js
```diff
@@ -499,58 +559,115 @@ class MergeContext {
...
+    // Staged commit needs all PR-required checks (configured on GitHub)
+    // so that GitHub could merge it into the protected base branch.
+    async _supplyStagingWithPrRequired(statusResult) {
+        assert(statusResult.succeeded());
+
+        const requiredContexts = await this._getRequiredContexts();
+        if (!requiredContexts)
+            return;
+
+        let requiredPRs = 0;
```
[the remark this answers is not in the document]
eduard-bagdasaryan: This variable is needed for the assert below. This assert is needed to control whether the staged commit has all GitHub-configured required status checks (the total number of staged commit checks can be more than this value)

src/MergeContext.js
```diff
@@ -499,58 +559,115 @@ class MergeContext {
...
+    }
+
+    // returns filled StatusResult object
+    async _getStagingStatuses() {
+        let combinedStatus = await GH.getStatuses(this._tagSha);
+        // TODO: use the assert below.
+        // The bot should be aware about all passed staging checks. We need to assert
+        // if we got more checks than expected (>Config.stagingChecks()). However,
+        // this is not possible in staged_run/guarded_run modes (after rerun), because
+        // we cannot separate staging extra checks, created by _supplyStagingWithPrRequired()
```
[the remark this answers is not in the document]
eduard-bagdasaryan: I think we can adjust the description, not changing status 'context' attributes.

src/MergeContext.js
```diff
@@ -499,58 +559,115 @@ class MergeContext {
...
-        // matching rule.
-        // For example, if we configured three required checks(Config.stagingChecks=3) but
-        // the branch has a single required check named 'Jenkins(build test)',
-        // the bot will wait for three checks with 'Jenkins(build test).*' names.
-        const needMatching = checksNumber !== undefined;
-        const requiredChecksNumber = needMatching ? checksNumber : requiredContexts.length;
-
-        // An array of [{context, state}] elements
-        let requiredChecks = [];
+            return null;
```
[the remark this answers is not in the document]
eduard-bagdasaryan: The caller can check with '!result' instead of '!result.length'. Not a great benefit, but a bit simpler. As I see, callers anyway do this check and return almost immediately.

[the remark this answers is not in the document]
rousskov: You may be missing my point: There should be no checks in the callers, not for null array, not for an empty array. If we return an empty array, the callers code should work correctly without any special logic dedicated to this case (because it is not so special). If we return `null`, we create a special case that must be treated specially in the callers.

src/MergeContext.js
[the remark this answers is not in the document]
rousskov: Why should approvalPending() return false when an approval is pending? Approvals have nothing to do with failed tests.

src/MergeContext.js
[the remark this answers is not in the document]
rousskov: That matches my expectations. Let's do that so that we do not need to deal with fake/copied status checks in StatusResult class and/or its users.

src/MergeContext.js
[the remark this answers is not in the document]
rousskov: I understand what you are saying, but I do not see the relevance. I still hope this variable will not be needed after other fixes, but if it stays, it needs to be renamed.

src/MergeContext.js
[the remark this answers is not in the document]
rousskov: Please note that my primary concern here is that the method code seems to be relying on the state of the "PR Approval" status check on GitHub. We should not rely on that. When needed, we should rely on our own, fresh calculation of the PR approval state (approved, suspended, etc.). If that concern is addressed by avoiding searching status checks for Config.approvalContext, then the method will become out of place -- it would no longer belong to the class dedicated to the analysis of those status checks because it would not be using those status checks anymore.

src/MergeContext.js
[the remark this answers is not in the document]
rousskov: > Eduard: commitStatus.approvalPending() should return false if a test has failed
> Alex: Why should approvalPending() return false when an approval is pending?
> Eduard: This statement is not correct

My statement is implied by yours. I am just asking why we should use such odd logic.

> approvalPending() returns true if and only if all passed tests succeeded except "PR approval" test which is pending.

Yes. The first condition looks unrelated to the method (triggering my question), and the second condition should probably be based on our approval calculation rather than GitHub status check (but that is a separate problem).

This code is gone in the latest revision, right? If so, we do not need to discuss it anymore.

src/MergeContext.js
```diff
@@ -499,58 +557,90 @@ class MergeContext {
...
+        assert(stagedStatuses.succeeded());
 
-        return ret ? 'success' : 'failure';
+        const requiredPrContexts = await this._getRequiredContexts();
+        const prStatuses = await this._getPRStatuses();
+
+        for (let requiredContext of requiredPrContexts) {
+            if (!stagedStatuses.requiredStatuses.some(el => el.context.trim() === requiredContext.trim())) {
+                const requiredPrStatus = prStatuses.requiredStatuses.find(el => el.context.trim() === requiredContext.trim());
+                assert(requiredPrStatus);
```
rousskov: Let's also assert that `requiredPrStatus.description` does not end with copiedDescriptionSuffix().

src/MergeContext.js
```diff
@@ -108,6 +108,82 @@ class Approval {
...
+        this.description = raw.description;
+
+        this._raw = raw;
+    }
+}
+
+// Passed required status checks analysis
+class StatusResult
+{
+    // checksNumber:
```
rousskov: Please s/checksNumber/expectedStatusCount/ to clarify its meaning (this is the minimum number of tests we expect to see) and to make its suffix more consistent with other Factory code that uses `count` rather than `number` (for English reasons: They say "number of apples" but "apple count").

src/MergeContext.js
```diff
@@ -108,6 +108,82 @@ class Approval {
...
+        let combinedStatus = "context: " + this.context + " expected/received: " + this.checksNumber + "/" +
+            this.requiredStatuses.length + ", combined: ";
+        if (this.failed())
+            combinedStatus += "failure";
+        else if (this.succeeded())
+            combinedStatus += "success";
+        else if (!this.final())
+            combinedStatus += "to-be-determined";
+        else
+            combinedStatus += "unexpected";
```
rousskov: I would reorder the last two and add a trailing comment to clarify the final() case: `// final() implies failed() or succeeded()`

src/MergeContext.js
```diff
@@ -108,6 +108,82 @@ class Approval {
...
+        this.state = raw.state;
+        this.targetUrl = raw.target_url;
+        this.description = raw.description;
+
+        this._raw = raw;
+    }
+}
+
+// Passed required status checks analysis
+class StatusResult
```
rousskov: s/StatusResult/StatusChecks/gi and describe the class as `aggregates (required) status check for a PR or commit`

src/MergeContext.js
```diff
@@ -499,58 +557,90 @@ class MergeContext {
...
+        }
+        return statusResult;
+    }
+
+    async _checkPRStatuses(logContext) {
+        const prStatus = await this._getPRStatuses();
+        this._log("status details: " + prStatus);
+        if (prStatus.succeeded())
+            return StepResult.Succeed();
+        // non-final or failed
```
rousskov: Please convert this comment into an assertion.

src/MergeContext.js
```diff
@@ -499,58 +557,90 @@ class MergeContext {
...
+        let statusResult = new StatusResult(requiredContexts.length, "PR");
+        // fill with required status checks
+        for (let st of combinedPrStatus.statuses) {
+            if (requiredContexts.some(el => el.trim() === st.context.trim() && el.trim() !== Config.approvalContext()))
+                statusResult.addRequiredStatus(new StatusCheck(st));
+        }
+        return statusResult;
+    }
+
+    async _checkPRStatuses(logContext) {
```
rousskov: s/getPRStatuses/getPrStatuses/ for readability sake

src/MergeContext.js
```diff
@@ -499,58 +557,90 @@ class MergeContext {
...
+        return StepResult.Fail();
+    }
+
+    // returns filled StatusResult object
+    async _getStagingStatuses() {
+        const combinedStagingStatuses = await GH.getStatuses(this._tagSha);
+        const genuineStatuses = combinedStagingStatuses.statuses.filter(st => !st.description.endsWith(Config.copiedDescriptionSuffix()));
+        assert(genuineStatuses.length <= Config.stagingChecks());
+        let statusResult = new StatusResult(Config.stagingChecks(), "Staging");
+        // all checks are 'required'
```
rousskov: Either remove this comment because it mismatches the code or s/all/all genuine/

src/MergeContext.js
```diff
@@ -499,58 +557,90 @@ class MergeContext {
...
+        return statusResult;
+    }
+
+    async _processStagingStatuses() {
+        const stagingStatus = await this._getStagingStatuses();
+        this._log("status details: " + stagingStatus);
+        if (stagingStatus.failed()) {
+            this._log("staging checks failed");
+            return await this._cleanupMergeFailed(false, this._labelFailedStagingChecks);
+        } else if (!stagingStatus.final()) {
```
rousskov: Please remove the `else` keyword to make this code simpler and more similar to what happens after this/second `if`.

src/MergeContext.js
```diff
@@ -499,58 +557,90 @@ class MergeContext {
...
+
+    // returns filled StatusResult object
+    async _getPRStatuses() {
+        let requiredContexts = await this._getRequiredContexts();
+        requiredContexts = requiredContexts.filter(el => el.trim() !== Config.approvalContext());
+        const combinedPrStatus = await GH.getStatuses(this._prHeadSha());
+        let statusResult = new StatusResult(requiredContexts.length, "PR");
+        // fill with required status checks
+        for (let st of combinedPrStatus.statuses) {
+            if (requiredContexts.some(el => el.trim() === st.context.trim() && el.trim() !== Config.approvalContext()))
```
rousskov: The "not approvalContext()" condition should be removed both because the "PR Approval" check is not really special and because that check is counted in requiredContexts.length.

rousskov: > This check is 'special' (at least for now) because this is the only check which may be delayed.

I see nothing special here. Virtually all checks take some time to go from pending to final. I suspect you attach some special meaning to the delay of this check because Anubis is responsible for ending the corresponding wait; for me, that responsibility lies outside the StatusResult scope.

None of the two proposed solutions make sense to me. What is wrong with:

* Remove any special handling of the "PR Approval" status check from the StatusResult code. For that class (and its object creators), the check is not special.
* (Continue to?) compute approval and approval-related waiting delays separately, based on the existing Approval class. That computation and related status check updates should happen _before_ StatusResult code gets status checks so that StatusResult gets a fresh "PR Approval" status check.

eduard-bagdasaryan: > (Continue to?) compute approval [...]

This is already done. Here is why (IMO) this does not work:

1. _checkMergeConditions() proceeds up to approval checking calculate them. E.g., it is 'slow burner' with a delay.
2. Since the approval is still pending, we update approval status check with fresh 'pending'.
3. Status check method starts (e.g., all other statuses are success, except PR approval, which is pending), so it returns pending.
4. _checkMergeConditions() notices that the aggregate status check is 'pending' and suspends the bot (instead of delaying with the interval calculated at (1)).

We could return 'delayed' after (2) (thus skipping status checking), but this approach is wrong either.

rousskov: > _checkMergeConditions() notices that the aggregate status check is 'pending' and suspends the bot (instead of delaying with the interval calculated at (1)).

How about changing this logic to suspend the bot only if the earlier approval check did not result in a delayed approval? Overall principle: We should never suspend (i.e., trigger an infinite wait) if there is change expected after a finite wait.

src/MergeContext.js
```diff
@@ -108,6 +108,84 @@ class Approval {
...
+        assert(raw.state);
+        assert(raw.target_url);
+        assert(raw.description);
+
+        this.context = raw.context;
+        this.state = raw.state;
+        this.targetUrl = raw.target_url;
+        this.description = raw.description;
+
+        this._raw = raw;
```
rousskov: Remove as unused? I was thinking "context", "state", etc. would become getters for "_raw" so that you can still return and use the raw object elsewhere when needed, but if you do not need the raw object, then there is no need to keep a reference to it, I guess.

src/MergeContext.js
```diff
@@ -108,6 +108,84 @@ class Approval {
...
+    }
+}
+
+// aggregates (required) status check for a PR or commit
```
rousskov: s/check/checks/

src/MergeContext.js
```diff
@@ -108,6 +108,84 @@ class Approval {
...
+
+    toString() {
+        let combinedStatus = "context: " + this.context + " expected/received: " + this.expectedStatusCount + "/" +
+            this.requiredStatuses.length + ", combined: ";
+        if (this.failed())
+            combinedStatus += "failure";
+        else if (this.succeeded())
+            combinedStatus += "success";
+        else if (this.final()) {
+            // final() implies failed() or succeeded()
```
rousskov: Let's move this comment so that it trails the assignment below it. That way, we can remove curly braces and all these chained `if` statements will look similar (as they should).

src/MergeContext.js

src/MergeContext.js

src/MergeContext.js

PR #18, opened 2018-05-17
PR message freshness postcondition

eduard-bagdasaryan:
With this additional postcondition, the bot tracks situations when PR
message was changed after tag creation. If so, the PR merging is started
from scratch, as usual.

### DIFF
--- a/src/MergeContext.js
+++ b/src/MergeContext.js
@@ -256,7 +256,7 @@ class MergeContext {
         return result;
     }
 
-    // Whether the PR message has not changed since the PR staged commit creation.
+    // Whether the commit message configuration remained intact since staging.
     async _messageIsFresh() {
         const tagCommit = await this._tagCommit();
         const result = this._prMessage() === tagCommit.message;
--- a/src/MergeContext.js
+++ b/src/MergeContext.js
@@ -118,6 +118,8 @@ class MergeContext {
         this._shaLimit = 6;
         // information used for approval test status creation/updating
         this._approval = null;
+        // optimization: cached _tagCommit() result
+        this._tagCommitCache = null;
     }
 
     // returns filled StepResult object
@@ -235,12 +237,18 @@ class MergeContext {
         return false;
     }
 
+    async _tagCommit() {
+        if (this._tagCommitCache)
+            return this._tagCommitCache;
+         this._tagCommitCache = await GH.getCommit(this._tagSha);
+         return this._tagCommitCache;
+    }
     // Whether the PR merge commit has not changed since the PR staged commit creation.
     // Note that it does not track possible conflicts between PR base branch and the
     // PR branch (the PR merge commit is recreated only when there are no conflicts).
     // Conflicts are tracked separately, by checking _prMergeable() flag.
     async _tagIsFresh() {
-        const tagCommit = await GH.getCommit(this._tagSha);
+        const tagCommit = await this._tagCommit();
         const prMergeSha = await GH.getReference(this._mergePath());
         const prCommit = await GH.getCommit(prMergeSha);
         const result = tagCommit.tree.sha === prCommit.tree.sha;
@@ -250,7 +258,7 @@ class MergeContext {
 
     // Whether the PR message has not changed since the PR staged commit creation.
     async _messageIsFresh() {
-        const tagCommit = await GH.getCommit(this._tagSha);
+        const tagCommit = await this._tagCommit();
         const result = this._prMessage() === tagCommit.message;
         this._log("tag message freshness: " + result);
         return result;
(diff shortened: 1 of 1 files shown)


src/MergeContext.js
```diff
@@ -262,6 +272,8 @@ class MergeContext {
     async _mayContinue() {
         if (!(await this._tagIsFresh()))
             return StepResult.Fail();
+        if (!(await this._messageIsFresh()))
```
rousskov: We also call _tagIsFresh() from _stagingFailed(). Should we call _messageIsFresh() from there as well?

src/MergeContext.js
```diff
@@ -243,7 +243,17 @@ class MergeContext {
         const tagCommit = await GH.getCommit(this._tagSha);
         const prMergeSha = await GH.getReference(this._mergePath());
         const prCommit = await GH.getCommit(prMergeSha);
-        return tagCommit.tree.sha === prCommit.tree.sha;
+        const result = tagCommit.tree.sha === prCommit.tree.sha;
+        this._log("tag freshness: " + result);
+        return result;
+    }
+
+    // Whether the PR message has not changed since the PR staged commit creation.
```
rousskov: Please use `Whether the commit message configuration remained intact since staging.`

src/MergeContext.js
eduard-bagdasaryan: I think we could but probably we should not. This _stagingFailed() is used on precondition checking: having a failed(but fresh) tag means that the code is unchanged and we need to skip this PR. The PR message is unrelated here (until however, some hypothetic checks would check PR messages also).

rousskov: I am interpreting your explanation as follows: We should not call _messageIsFresh() from _stagingFailed() because _stagingFailed() is dedicated to checking the state of the tag and the commit message has nothing to do with the tag.

I certainly agree with that logic, but the answer misses the essence of my concern. Unfortunately, I was misled by the _stagingFailed() name when formulating my concern. Please allow me to rephrase using a better term from your answer:

* Should _messageIsFresh() become a "precondition", just like _tagIsFresh() is a "precondition"?

src/MergeContext.js
```diff
@@ -235,12 +237,18 @@ class MergeContext {
         return false;
     }
 
+    async _tagCommit() {
+        if (this._tagCommitCache)
+            return this._tagCommitCache;
+         this._tagCommitCache = await GH.getCommit(this._tagSha);
```
rousskov: You can simplify by making the assignment (i.e., cache initialization) conditional instead of returning twice.

src/MergeContext.js
rousskov: > _tagIsFresh() itself is not a precondtion, the precondition is _stagingFailed() which has more 'narrow' meaning

Not sure why you are making it so difficult, but I guess I have to rephrase again:

* Should _messageIsFresh() become one of the conditions that may be evaluated whenever _stagingFailed() or similar conditions are evaluated?

> If the tag is all information associated with the staged commit (including the 'tree' object of the 'merge' commit and the commit message)

AFAIK, the tag in this context is a (named) staged commit.

Please note that I am not (yet) discussing where exactly a second _messageIsFresh() check should be added. I am just asking whether we should add a second _messageIsFresh() check somewhere where other, similar conditions are already evaluated.

> If we decide that tag becomes stale when its commit message is stale(compared to the PR message), then yes

IMHO, at the current level of abstraction/detail, the tag is indistinguishable from the staged commit. They should be thought of as the same thing. The tag is a name of (or a reference pointing to) the staged commit. Needless to say, _tagIsFresh() and/or other methods use different levels of abstraction; at some point the difference between a name and the object being named becomes important, but we should not care about that in this discussion.

What you might be implying is that "staged commit freshness" includes several concepts that take things like the master branch state, the PR branch state, and/or the GitHub merge commit state into account. Some of them may be "OK/fresh" while others may be "requiring an update/stale". _stagingFailed() evaluates some but possibly not all such concepts. I do not think any of that is important for now.

If we define "fresh" as "all of the components are fresh" and if we assume that a change in the PR description make the staged commit "stale", should we add another check for the PR description somewhere?

> for simplicity

Other than the increased complexity of adding another check to the code, what are the trade-offs? That is, what will (not) happen if we add or do not add that second check?

rousskov: > With this change, if [...] somebody modified the PR description, the PR will be processed (and tag recreated).

Let's recreate the staging commit when the PR description has changed.

Disclaimer: I suspect that when you say "proceed", "tag failed (checks failed)", "skipped", and "processed", you are relying on some Anubis code-specific terminology that I am not very familiar with. The combination of those terms in your last paragraph confuses me a lot because they seem to imply something that does not happen (or should not happen; or should not matter whether it happens) in reality. I hope my interpretation of what you are saying is correct, but I cannot be sure. I am reasonably sure that we should be recreating staging commits that no longer match the PR, and the PR description is one of those factors we should be monitoring.

PR #23, opened 2018-05-31
Split huge all-purpose MergeContext into several classes

eduard-bagdasaryan:
After this refactoring, two new classes were introduced:

* MergeInitiator: starts PR prorcessing up to staged commit creation
* MergeFinalizer: finishes processing a staged PR

### DIFF
--- a/src/MergeContext.js
+++ b/src/MergeContext.js
@@ -546,8 +546,7 @@ class MergeContext {
 } // MergeContext
 
 // Starts PR processing.
-// Performs different checks over the PR and creates staged commit
-// for it.
+// Performs all PR merge steps up to and including the staged commit creation.
 class MergeInitiator extends MergeContext {
 
     constructor(pr) {
@@ -709,9 +708,8 @@ class MergeInitiator extends MergeContext {
 } // MergeInitiator
 
 // Finishes PR processing.
-// Waits for staging checks completing, merges staged commit
-// into base branch and closes the PR. Operates only on PRs having
-// staged commit and staging branch pointing to it.
+// Performs all PR merge steps after the staged commit creation.
+// Requires PRs having staged commit and staging branch pointing to it.
 class MergeFinalizer extends MergeContext {
     constructor(pr, tSha) {
         super(pr);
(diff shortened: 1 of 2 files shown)


src/MergeContext.js
```diff
@@ -856,10 +513,302 @@ class MergeContext {
...
+    _toString() {
+        let str = this._debugString();
+        if (this._tagSha !== null)
+            str += ", tag: " + this._tagSha.substr(0, this._shaLimit);
+        return str + ")";
+    }
+} // MergeContext
+
+// Starts PR processing.
+// Performs different checks over the PR and creates staged commit
```
rousskov: I would generalize the second sentence into something like "Performs all PR merge steps up to and including the staged commit creation".

src/MergeContext.js
```diff
@@ -856,10 +513,302 @@ class MergeContext {
...
+        await this._unlabelPreconditionsChecked();
+        await this._createStaged();
+        await this._labelWaitingStagingChecks();
+
+        return StepResult.Succeed();
+    }
+} // MergeInitiator
+
+// Finishes PR processing.
+// Waits for staging checks completing, merges staged commit
```
rousskov: I would not try to enumerate the steps here -- the list will quickly become stale. I would generalize the second sentence into something like "Performs all PR merge steps after the staged commit creation".

src/MergeContext.js
```diff
@@ -856,10 +513,302 @@ class MergeContext {
...
+// into base branch and closes the PR. Operates only on PRs having
+// staged commit and staging branch pointing to it.
```
rousskov: If this class checks that there is a staging branch pointing to the staged commit, then this should not be documented here because it is just one of the many checks/things this class does.

Otherwise (i.e., if somebody outside must ensure that the staging branch points to the staged commit), s/Operates only on/Requires/ but also consider moving that enforcement inside this class.

src/MergeContext.js

src/MergeContext.js

src/MergeContext.js

PR #24, opened 2018-05-31
guarded_run: M-cleared-for-merge PRs should not wait for others

eduard-bagdasaryan:
In guarded_run mode, PRs with the M-cleared-for-merge label have the
highest priority. The bot now gives up a being-processed unlabeled PR if
there is at least one labeled PR.

### DIFF
--- a/src/PrMerger.js
+++ b/src/PrMerger.js
@@ -28,17 +28,16 @@ class PrMerger {
         let prList = await GH.getPRList();
         for (let pr of prList) {
             pr.clearedForMerge = await this._clearedForMerge(pr.number);
-            if (finalizer !== null && finalizer.number() === pr.number) {
-                pr.isCurrent = true;
-                pr.processor = finalizer;
-            } else {
-                pr.isCurrent = false;
-                pr.processor = new MergeInitiator(pr);
-            }
+            if (finalizer && finalizer.prNumber() === pr.number)
+                pr.anubisProcessor = finalizer;
+            else
+                pr.anubisProcessor = new MergeInitiator(pr);
         }
 
         prList.sort((pr1, pr2) => { return (Config.guardedRun() && (pr2.clearedForMerge - pr1.clearedForMerge)) ||
-                pr2.isCurrent - pr1.isCurrent || pr1.number - pr2.number; });
+                pr2.anubisProcessor.isFinalizer() - pr1.anubisProcessor.isFinalizer() ||
+                pr1.number - pr2.number;
+        });
         this._logPRList(prList);
         return prList;
     }
@@ -58,11 +57,14 @@ class PrMerger {
         const finalizer = await this._current();
         const prList = await this._getPRList(finalizer);
 
+        this.total = 0;
         while (prList.length) {
             try {
                 const pr = prList.shift();
-                if (await this._process(pr))
-                    return true; // still in process
+                this.total++;
+                const result = await pr.anubisProcessor.process();
+                if (!this.prDone(pr.anubisProcessor, result))
+                    return true;
             } catch (e) {
                 this.errors++;
                 if (prList.length)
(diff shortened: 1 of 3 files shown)


src/PrMerger.js
```diff
@@ -43,12 +44,31 @@ class PrMerger {
...
+        let currentContext = await this._current();
+
+        if (currentContext && Config.guardedRun()) {
+            const clearedForMerge = await this._clearedForMerge(currentContext.number());
+            if (!clearedForMerge && prList.length && prList[0].clearedForMerge) {
+                Logger.info("will try to give up the not-cleared-for-merge PR" + currentContext.number() +
+                        " in favor of a cleared-for-merge PR" + prList[0].number);
+                // Give only other PRs a chance to start.
+                // Without removing current PR from the list, we can get into
+                // a loop of useless restarting it from scratch.
```
rousskov: If we can get into a loop, then there is a bug somewhere that we should fix instead AFAICT. The situation is simple: When there are labeled PRs, we abort any current merging of an unlabeled PR and do not start merging of any unlabeled PR.

If it makes things easier, we do not have to start merging any unlabeled PR _regardless_ of whether labeled PRs exist or not. In other words, we only start merging labeled PRs.

All of the above applies to guarded runs only, of course.

src/PrMerger.js
```diff
@@ -43,12 +44,31 @@ class PrMerger {
...
+                prList = prList.filter(pr => pr.number !== currentContext.number());
+                currentContext = null;
+            }
+        }
+
+        if (currentContext) {
+            if (!(await this._finishContext(currentContext)))
+                return true; // still in-process
+            // refresh the list
+            prList = await this._getPRList();
```
rousskov: I do not like the complexity this "cache update" introduces. After we finished merged something, can we just return and start from scratch instead?

src/PrMerger.js
```diff
@@ -109,7 +129,11 @@ class PrMerger {
         }
         const prNum = Util.ParseTag(tag.ref);
         Logger.info("PR" + prNum + " is the current");
+        // TODO: handle 'not found' errors
         const stagingPr = await GH.getPR(prNum, false);
+        assert(stagingPr);
+        stagingPr.clearedForMerge = await this._clearedForMerge(prNum);
```
rousskov: Sooner or later, we will forget to initialize this member. Let's initialize it in the PR constructor and only there. AFAICT, such initialization is only needed in the guarded run mode.

src/PrMerger.js
```diff
@@ -43,12 +44,31 @@ class PrMerger {
     // there is a PR still-in-merge.
     async runStep() {
         Logger.info("runStep running");
-        const currentContext = await this._current();
-        if (currentContext && !(await this._finishContext(currentContext)))
```
rousskov: Does _finishContext() _start_ something long/expensive? If it only makes a few GitHub calls without initiating anything, then I think it is OK to let it run here. We just cannot `return true` if the currentContext is unlabeled and there are labeled contexts now. On other other hand, we do not really want to `return true` in any case because other PRs may need some processing too, even of all of them are unlabeled.

src/PrMerger.js
eduard-bagdasaryan: > If we can get into a loop

We could get into the loop without my changes. With my solution, the bot switches from the unlabeled PR to a labeled only if such labeled PR has passed all checks because there can be many labeled, but defective PRs. If all labeled PRs are invalid yet, the bot keeps hanging on the unlabeled PR (and waiting until it becomes labeled).
With your (1) solution we would delete staged commit for our unlabeled PR and would not start any other PR. With your (2) solution we would not even start the unlabeled PR at all.
I think my solution is better because it works for more cases (bot processes more PRs) while it does not have disadvantages, IMO.

src/PrMerger.js
```diff
@@ -43,12 +44,31 @@ class PrMerger {
...
-        const prList = await this._getPRList();
-        this._logPRList(prList);
+        let prList = await this._getPRList();
+
+        let currentContext = await this._current();
+
+        if (currentContext && Config.guardedRun()) {
+            const clearedForMerge = await this._clearedForMerge(currentContext.number());
+            if (!clearedForMerge && prList.length && prList[0].clearedForMerge) {
+                Logger.info("will try to give up the not-cleared-for-merge PR" + currentContext.number() +
```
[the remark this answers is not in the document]
eduard-bagdasaryan: > Why such an uncertainty?

This uncertainty is needed because there is a chance when this attempt (of giving up) was not successful and we are still working with the unlabeled PR. This situation can happen when all labeled PRs are defective.

> It do not think this code should know about labeled PR processing order.

OK.

src/PrMerger.js
eduard-bagdasaryan: > Does _finishContext() start something long/expensive?

The bot does not have expensive methods: GitHub calls are not expensive AFAICT. This method (as well as another, startProcessing()), initiate several (or several dozens) API calls (taking several seconds).
So, I can try to follow this suggestion (simplifying the code at the expense of optimization).

src/PrMerger.js
eduard-bagdasaryan: We could "start from scratch" but this would make another complexity: how the upper-level code (RepoMerger::run()) would know that we want to rerun? We would have to manipulate with RepoMerger::rerun flag in low-level code, which is not good (IIRC we rejected such manipulation some time ago).

src/PrMerger.js
rousskov: > Does _finishContext() start something long/expensive?

For example, something that may trigger new CI tests.

eduard-bagdasaryan:
@rousskov, I reworked this PR, please review. The major change is in 3557cd.
72caf69 is out of scope (it is a fix for recently merged PR23 into master).

src/PrMerger.js
```diff
@@ -24,12 +24,22 @@ class PrMerger {
...
+                pr.processor = finalizer;
+            } else {
+                pr.isCurrent = false;
+                pr.processor = new MergeInitiator(pr);
+            }
+        }
 
-        prList.sort((pr1, pr2) => { return pr2.clearedForMerge - pr1.clearedForMerge || pr1.number - pr2.number; });
+        prList.sort((pr1, pr2) => { return (Config.guardedRun() && (pr2.clearedForMerge - pr1.clearedForMerge)) ||
+                pr2.isCurrent - pr1.isCurrent || pr1.number - pr2.number; });
```
rousskov: I would move the third (pr.number) condition to its own line, to emphasize that we have three sorting criteria.

PR #25, opened 2018-06-11
Keep PR GitHub metadata up-to-date

eduard-bagdasaryan:
When reacting to an event, advance each open PR as much as possible instead of
waiting for the being-merged PR to fully merge before advancing the next PR.
At the minimum, update PR labels.

Keeping all PRs up-to-date required many architectural and implementation
fixes. Most of the "middle" code (somewhere between highest-level GitHub event
processing code and low-level GitHub operations) was rewritten.

Also remove stale PR tags of closed PRs.

Also fixed GH.getTags() which unexpectedly threw when getting an
empty tag list from GitHub.

### DIFF
--- a/src/MergeContext.js
+++ b/src/MergeContext.js
@@ -166,11 +166,13 @@ class StatusChecks
 
     addRequiredStatus(requiredStatus) {
         assert(requiredStatus);
+        assert(!this.hasStatus(requiredStatus.context));
         this.requiredStatuses.push(requiredStatus);
     }
 
     addOptionalStatus(optionalStatus) {
         assert(optionalStatus);
+        assert(!this.hasStatus(optionalStatus.context));
         this.optionalStatuses.push(optionalStatus);
     }
 
@@ -186,7 +188,8 @@ class StatusChecks
                 el.description === approval.description);
     }
 
-    addApprovalStatus(approval) {
+    setApprovalStatus(approval) {
+        this.requiredStatuses = this.requiredStatuses.filter(st => st.context !== Config.approvalContext());
         let raw = {
             state: approval.state,
             target_url: Config.approvalUrl(),
@@ -567,12 +570,12 @@ class PullRequest {
 
         if (!this._prStatuses.hasApprovalStatus(this._approval)) {
             await this._createApprovalStatus(this._prHeadSha());
-            this._prStatuses.addApprovalStatus(this._approval);
+            this._prStatuses.setApprovalStatus(this._approval);
         }
 
         if (this._stagedStatuses && !this._stagedStatuses.hasApprovalStatus(this._approval)) {
             await this._createApprovalStatus(this._tagSha);
-            this._stagedStatuses.addApprovalStatus(this._approval);
+            this._stagedStatuses.setApprovalStatus(this._approval);
         }
     }
 
--- a/src/MergeContext.js
+++ b/src/MergeContext.js
@@ -866,6 +866,8 @@ class PullRequest {
         Log.LogException(e, this._toString() + ' ' + msg);
     }
 
+    // TODO: consider moving this and other similar checks
+    // directly into GH methods
     // TODO: Rename to _readOnly()
     // whether all GitHub/repository changes are prohibited
     _dryRun(msg) {
--- a/src/MergeContext.js
+++ b/src/MergeContext.js
@@ -365,23 +365,6 @@ class PrState
     }
 }
 
-// external limits for PullRequest::process() freedoms
-class PrRestrictions
-{
-    constructor() {
-        this._banStaging = false; // do not stage; only valid for brewing PRs
-    }
-
-    stagingBanned() {
-        return this._banStaging;
-    }
-
-    banStaging(bool) {
-        this._banStaging = bool;
-        return this;
-    }
-}
-
 // Determines the feature branch position relative to its base branch: ahead,
 // merged, or diverged. This class essentially decides whether/which one of
 // the two branches is fully contained in another.
@@ -425,9 +408,7 @@ class BranchPosition
 // a single GitHub pull request
 class PullRequest {
 
-    constructor(pr, restrictions) {
-        assert(restrictions instanceof PrRestrictions);
-
+    constructor(pr, banStaging) {
         this._rawPr = pr; // may be rather old and lack pr.mergeable; see _loadRawPr()
 
         this._shaLimit = 6; // how many SHA chars to show in debug messages
@@ -452,7 +433,7 @@ class PullRequest {
         this._prState = null; // calculated PrState
 
         this._labels = null;
-        this._restrictions = restrictions;
+        this._stagingBanned = banStaging; // do not stage; only valid for brewing PRs
         this._updated = false; // _update() has been called
 
         // truthy value contains a reason for disabling _pushLabelsToGitHub()
@@ -700,7 +681,7 @@ class PullRequest {
         if (this._approval.grantedTimeout())
             throw this._exSuspend("waiting for objections");
 
-        if (this._restrictions.stagingBanned())
+        if (this._stagingBanned)
             throw this._exSuspend("waiting for another staged PR");
 
         if (this._stagedCommitWillFail)
@@ -1053,7 +1034,7 @@ class PullRequest {
         if (this._stagingOnly())
             throw this._exSuspend("waiting for staging-only mode to end");
 
-        assert(!this._restrictions.stagingBanned());
+        assert(!this._stagingBanned);
 
         try {
             await GH.updateReference(this._prBaseBranchPath(), this._tagSha, false);
@@ -1108,7 +1089,7 @@ class PullRequest {
         await this._stagedPosition.compute();
         assert(this._stagedPosition.ahead());
 
-        assert(!this._restrictions.stagingBanned());
+        assert(!this._stagingBanned);
         await GH.updateReference(Config.stagingBranchPath(), this._tagSha, true);
         this._prState = PrState.Staged();
     }
@@ -1263,9 +1244,7 @@ class PullRequest {
 
 // PullRequest::process() wrapper
 async function Process(rawPr, banStaging) {
-    let restrictions = new PrRestrictions().banStaging(banStaging);
-
-    let pr = new PullRequest(rawPr, restrictions);
+    let pr = new PullRequest(rawPr, banStaging);
     let delayMs = await pr.process();
 
     let result = new ProcessResult();
--- a/src/MergeContext.js
+++ b/src/MergeContext.js
@@ -431,7 +431,8 @@ class PullRequest {
         this._prState = null; // calculated PrState
 
         this._labels = null;
-        this._stagingBanned = banStaging; // do not stage; only valid for brewing PRs
+        // while unexpected, PR merging and closing is not prohibited when staging is
+        this._stagingBanned = banStaging;
         this._updated = false; // _update() has been called
 
         // truthy value contains a reason for disabling _pushLabelsToGitHub()
@@ -599,17 +600,18 @@ class PullRequest {
         return this._tagCommitCache;
     }
 
-    // Returns true iff:
-    // * PR staged commit is ahead of the base branch and
-    // * the PR staged commit tree is equal to the PR merge commit tree and
-    // * the PR staged commit message is equal to the PR description on GitHub.
-    // Note that since PR merge commit is automatically re-created by GitHub
-    // only if the PR branch has no conflicts with the base branch, it is not
-    // possible to track such conflicts merely comparing commits.
-    // Conflicts are tracked separately, by checking _prMergeable() flag.
-    async _tagIsFresh() {
+    // Determines whether the existing staged commit is equivalent to a staged
+    // commit that could be created right now. Relies on PR merge commit being fresh.
+    // TODO: lacks some checks, e.g., mismatching staging checks list (configured
+    // on GitHub) or staging commit message (Anubis now may generate a different message).
+    async _stagedCommitIsFresh() {
         assert(this._tagSha);
         if (!this._stagedPosition.ahead())
+            return false;
+
+        // check this separately because GitHub does not recreate PR merge commits
+        // for conflicted PR branches (leaving stale PR merge commits).
+        if (!this._prMergeable())
             return false;
 
         if (!(await this._messageIsFresh()))
@@ -869,7 +871,7 @@ class PullRequest {
         }
 
         // The staged commit became out of sync with PR and(or) base branches.
-        if (!(await this._tagIsFresh())) {
+        if (!(await this._stagedCommitIsFresh())) {
             this._log("the staged commit became stale due to PR branch and(or) base branch changes");
             this._prState = PrState.Brewing();
             return;
@@ -915,7 +917,7 @@ class PullRequest {
     async _messageIsFresh() {
         const tagCommit = await this._tagCommit();
         const result = this._prMessage() === tagCommit.message;
-        this._log("tag message freshness: " + result);
+        this._log("staged commit message freshness: " + result);
         return result;
     }
 
@@ -1132,7 +1134,7 @@ class PullRequest {
          */
         await this._loadTag(); // requires this._rawPr.number
         await this._loadRawPr(); // requires this._loadTag()
-        await this._loadPrState(); // requires this._loadPr()
+        await this._loadPrState(); // requires this._loadRawPr()
         this._log("PR state: " + this._prState);
         await this._loadLabels();
 
--- a/src/MergeContext.js
+++ b/src/MergeContext.js
@@ -522,7 +522,7 @@ class PullRequest {
             combinedStatus.statuses.find(el => el.context.trim() === Config.approvalContext()) : null;
 
         if (approvalStatus && this._approval.matchesGitHubStatusCheck(approvalStatus)) {
-            this._log("Approval status already exists: " + Config.approvalContext() + ", " + this._approval);
+            this._log("approval status already exists: " + Config.approvalContext() + ", " + this._approval);
             return;
         }
 
--- a/src/MergeContext.js
+++ b/src/MergeContext.js
@@ -139,6 +139,12 @@ class StatusCheck
         this.targetUrl = raw.target_url;
         this.description = raw.description;
     }
+
+    failed() { return !(this.pending() || this.success()); }
+
+    success() { return this.state === 'success'; }
+
+    pending() { return this.state === 'pending'; }
 }
 
 // aggregates status checks for a PR or commit
@@ -176,17 +182,23 @@ class StatusChecks
     // no more required status changes or additions are expected
     final() {
         return (this.requiredStatuses.length >= this.expectedStatusCount) &&
-            this.requiredStatuses.every(check => check.state !== 'pending');
+            this.requiredStatuses.every(check => !check.pending());
     }
 
     // something went wrong with at least one of the required status checks
     failed() {
-        return this.requiredStatuses.some(check => check.state !== 'pending' && check.state !== 'success');
+        return this.requiredStatuses.some(check => check.failed());
+    }
+
+    // same as failed(), but skips the given context when searching
+    failedExcept(context) {
+        const filteredChecks = this.requiredStatuses.filter(st => st.context !== context);
+        return filteredChecks.some(check => check.failed());
     }
 
     // the results are final and all checks were a success
     succeeded() {
-        return this.final() && this.requiredStatuses.every(check => check.state === 'success');
+        return this.final() && this.requiredStatuses.every(check => check.success());
     }
 
     toString() {
@@ -882,7 +894,7 @@ class PullRequest {
         const stagingStatuses = await this._getStagingStatuses();
         // Do not vainly recreate staged commit which will definitely fail again,
         // since the PR+base code is yet unchanged and the existing errors still persist
-        if (stagingStatuses.failed()) {
+        if (stagingStatuses.failedExcept(Config.approvalContext())) {
             this._freshStagedCommitWithFailedChecks = true;
             this._prState = PrState.Brewing();
             return;
@@ -925,8 +937,11 @@ class PullRequest {
         const stagingStatus = await this._getStagingStatuses();
         this._log("staging status details: " + stagingStatus);
 
-        if (stagingStatus.failed())
+        if (stagingStatus.failedExcept(Config.approvalContext()))
             throw this._exLabeledFailure("staging tests failed", Config.failedStagingChecksLabel());
+
+        // should have checked for approval already
+        assert(!stagingStatus.failed());
 
         if (!stagingStatus.final()) {
             this._labels.add(Config.waitingStagingChecksLabel());
@@ -1012,6 +1027,7 @@ class PullRequest {
             throw this._exObviousFailure("restart waiting for objections");
 
         const statusChecks = await this._getPrStatuses();
+        // the approval status have been checked already
         if (statusChecks.failed())
             throw this._exObviousFailure("new PR branch tests appeared/failed after staging");
 
--- a/src/MergeContext.js
+++ b/src/MergeContext.js
@@ -926,6 +926,12 @@ class PullRequest {
             return;
         }
 
+        // TODO: It feels wrong than an external restriction affects this PR
+        // state. Yes, this PR cannot be staged, but that does not mean (from
+        // this state-computing code point of view) that this PR is not, say,
+        // PrState.Staged(). I suspect this TODO will be easier to address
+        // after addressing the "Why are we modifying the repository in a
+        // load() method?!" TODO below.
         if (this._restrictions.stagingBanned()) {
             this._prState = PrState.Brewing();
             return;
@@ -972,6 +978,12 @@ class PullRequest {
 
         if (stagingStatus.failed()) {
             this._labels.add(Config.failedStagingChecksLabel());
+            // keep the staged commit tagged to avoid ...(XXX). TODO: It is
+            // not clear what general problems we can solve by keeping this PR
+            // tagged while allowing staging of other PRs. When the next PR
+            // gets staged, this PR will enter brewing state (by that state
+            // definition), and we will be exactly where we would be if we
+            // simply unstage here and now.
             throw this._exSuspendedFailure("staging tests failed");
         }
 
@@ -1098,6 +1110,8 @@ class PullRequest {
         } catch (e) {
             if (e.name === 'ErrorContext' && e.unprocessable()) {
                 await this._stagedPosition().compute();
+                // TODO: Test whether the staged commit is now a part of
+                // the base branch. .deverged() is not checking exactly that!
                 if (this._stagedPosition.diverged())
                     throw new Error("failed to fast-forward to the staged commit");
             }
@@ -1220,9 +1234,23 @@ class PullRequest {
             const unstageRequested = !suspended; // see exception table description
             // Do we need to delete the staging tag?
             // brewing: staging tag does not exist
-            // staged: get rid of the failed staging tag (if was requested)
+            // staged: get rid of the failed staging tag (unless keepStagedRequested())
             // merged: keep the staging tag for pending cleanup
+
+            // XXX: A brewing PR may still have a staging tag -- we adjust PR
+            // state for many reasons, some not related to tag existence. We
+            // should delete that stale tag here (by default; if any).
+
+            // XXX: Deleting the tag of a failed merged PR will result in that
+            // PR becoming "brewing" on the next scan, and an attempt to merge
+            // the same changes again. That is why we must never untag merged
+            // PRs after _finalize() failures AFAICT (and not because we want
+            // to wait for the pending cleanup that does not even apply to
+            // open merged PRs IIRC).
             if (this._prState.staged() && unstageRequested) {
+                // TODO: Also reset this._prState to PrState.Brewing().
+                // TODO: All this code should probably be moved to _unstage()
+                // which should call _untag() if brewering PRs need untagging.
                 this._removeStagingLabels();
                 if (!this._dryRun("cleanup failed staging tag"))
                     await GH.deleteReference(this._stagingTag())
@@ -1232,7 +1260,10 @@ class PullRequest {
                                 });
             }
 
-            // drop staged state and give way to others
+            // Drop staged state and give way to others. This is an ugly hack:
+            // We should not be lying about PR state or even know that our PR
+            // state creates a staging ban for other PRs.
+            // XXX: e may not be knownProblem and, hence, may not have dropStagingBanRequested()
             if (this._prState.staged() && e.dropStagingBanRequested())
                 this._prState = PrState.Brewing();
 
@@ -1260,7 +1291,7 @@ class PullRequest {
         // final (set after the PR is merged): Config.mergedLabel()
     }
 
-    // remove staging-related labels
+    // remove labels that only make sense for a staged PR
     _removeStagingLabels() {
         this._labels.remove(Config.failedStagingChecksLabel());
         this._labels.remove(Config.passedStagingChecksLabel());
--- a/src/MergeContext.js
+++ b/src/MergeContext.js
@@ -842,9 +842,9 @@ class PullRequest {
     staged() { return this._prState.staged(); }
 
     _debugString() {
-        const staged = this._tagSha ? "staged: " + this._tagSha.substr(0, this._shaLimit) + ' ' : "";
+        const tagged = this._tagSha ? "tagged: " + this._tagSha.substr(0, this._shaLimit) + ' ' : "";
         const detail =
-            "head: " + this._rawPr.head.sha.substr(0, this._shaLimit) + ' ' + staged +
+            "head: " + this._rawPr.head.sha.substr(0, this._shaLimit) + ' ' + tagged +
             "history: " + this._breadcrumbs.join();
         return "PR" + this._rawPr.number + ` (${detail})`;
     }
@@ -1245,6 +1245,7 @@ class PullRequest {
     // instant retries does not require reprocessing from scratch
     _exRetryNow(why) {
         assert(arguments.length === 1);
+        this._labelPushBan = why;
         this._reprocessingDelayMs = 0;
         return new PrProblem(why);
     }
--- a/src/MergeContext.js
+++ b/src/MergeContext.js
@@ -1164,7 +1164,7 @@ class PullRequest {
 
             // (by default) get rid of the failed staging tag (if any)
             if (!(knownProblem && e.keepStagedRequested()) &&
-                !this._prState.brewing() &&
+                this._prState.staged() &&
                 !this._dryRun("cleanup failed staging tag")) {
                 await GH.deleteReference(this._stagingTag())
                     .catch(deleteReferenceError => {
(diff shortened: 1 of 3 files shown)


src/MergeContext.js
```diff
@@ -377,15 +381,131 @@ class MergeContext {
...
+               Log.LogException(e, this._toString() + " " + this._stagingTag() + " not found");
+           else
+               throw e;
+       }
+    }
+
+    // Preconditions
+
+    // Check 'staging tag' state as merge condition.
+    // Returns true if there is a fresh merge commit with 'failure' status.
```
rousskov: s/merge commit with 'failure' status/staged commit with failed status checks/

src/MergeContext.js
```diff
@@ -377,15 +381,131 @@ class MergeContext {
...
+            this._logFailedCondition("not in progress");
+            return StepResult.Fail();
+        }
+
+        if (!this._prMergeable()) {
+            this._logFailedCondition("mergeable");
+            return StepResult.Fail();
+        }
+
+        if (await this._stagingFailed()) {
```
rousskov: This is not a *pre*condition AFAICT -- staged PRs are handled _after_ they are staged (naturally).

src/MergeContext.js
```diff
@@ -883,11 +837,6 @@ class MergeFinalizer extends MergeContext {
         if (!(await this._messageIsFresh()))
             return await this._cleanupMergeFailed(true, this._labelCleanStaged);
 
-        this._approval = await this._checkApproval();
-        this._log("checkApproval: " + this._approval);
-        await this._setApprovalStatus(this._prHeadSha());
-        await this._setApprovalStatus(this._tagSha);
-
         if (this._prInProgress()) {
```
rousskov: Some conditions, like this one (if it is interpreted without peeking at its current implementation) and _prOpen() are both pre- and post-conditions. Have you tried grouping them into a dedicated _checkTimelessConditions() method that the other two _check*conditions() methods would call?

src/MergeContext.js
```diff
@@ -938,9 +887,62 @@ class MergeFinalizer extends MergeContext {
...
+        }
+        assert(Config.githubUserEmail());
+
+        const user = await GH.getUser(Config.githubUserLogin());
+        Config.githubUserName(user.name);
+        assert(Config.githubUserName());
+    }
+
+    async _createStaged() {
+        this._log("start merging...");
```
rousskov: s/merging/staging/

src/PrMerger.js
```diff
@@ -54,17 +56,21 @@ class PrMerger {
...
-                const pr = prList.shift();
+                const rawPr = prList.shift();
                 this.total++;
-                const result = await pr.anubisProcessor.process();
-                if (!this.prDone(pr.anubisProcessor, result))
-                    return true;
+                let pr = new PullRequest(rawPr);
+                const result = await pr.process(merging);
+                if (!merging)
+                    merging = result.succeeded();
```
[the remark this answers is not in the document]
eduard-bagdasaryan: Done.

src/MergeContext.js
```diff
@@ -952,10 +954,40 @@ class MergeFinalizer extends MergeContext {
         return await this._cleanupMerged();
 
     }
-} // MergeFinalizer
+
+    // StepResult.Succeed: this PR is in-progress
+    // StepResult.Delay: this PR is delayed
```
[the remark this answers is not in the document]

src/MergeContext.js
rousskov: I do not understand the parts of your your response talking about _prInProgress() -- AFAICT that method checks the `WIP:` prefix. Checking that PR title prefix has nothing to do with tracking already merged PRs. While it is currently recorded as a part of the title, the prefix should be treated like a label. I suggest s/InProgress/WorkInProgress/ or even s/prInProgress/wipPr/ to reduce clashes between "PR merging progress" and "work-in-progress PR".

src/MergeContext.js
```diff
@@ -938,9 +941,61 @@ class MergeFinalizer extends MergeContext {
         }
     }
 
+    // Start processing
```
rousskov: Since "processing" can be interpreted in many ways (metadata updates, staging, merging, staging+merging), I suggest removing these "Start processing" and "Finish processing" labels. Also, grouping methods together like this rarely works long-term. Sooner or later, some of the "start processing" methods will be added elsewhere.

src/RepoMerger.js
```diff
@@ -96,6 +96,8 @@ class RepoMerger {
 
     _plan(ms) {
         assert(ms > 0);
+        const maxMs = Math.pow(2, 31) - 1;
```
rousskov: I would add a comment to clarify why we are restricting this value. For example, `obey nodejs setTimeout() limits`.

src/MergeContext.js
```diff
@@ -377,124 +516,224 @@ class MergeContext {
...
 
-    async _unlabelPreconditionsChecking() {
-        await this._removeLabelsIf([
-                Config.passedStagingChecksLabel(),
-                Config.waitingStagingChecksLabel()
-                ]);
+    // Atomize PR post-merge cleanup: suspend this PR
+    // until all required operations are complete.
+    async _finalize() {
+        if (!this._prState.postStaged() || this._dryRun("finalize"))
```
rousskov: If this method is not for other states, then we should assert postStaged() instead of silently suspending a PR.

src/MergeContext.js
```diff
@@ -377,124 +516,224 @@ class MergeContext {
...
+        try {
+            await this._applyLabels();
+            await GH.updatePR(this._prNumber(), 'closed');
+            if (this._prState.postStaged())
+                await GH.deleteReference(this._stagingTag());
+            this._log("finalize completed");
+            return StepResult.Succeed();
+        } catch (e) {
+            Log.LogException(e, this._toString() + " unexpected error while finalizing.");
+            return StepResult.Suspend();
```
rousskov: Why suspend a PR that fails cleanup? In the commit message, you said

> For post-staged exceptions, the PR is considered as 'suspended', and does not start another PR, until all required cleanup is completed.

Yes, I can see that happening in the code, but I do not know the answer to the "why" question.

src/MergeContext.js
```diff
@@ -855,107 +960,172 @@ class MergeFinalizer extends MergeContext {
...
+    // If possible, also merge or advance the PR towards merging.
+    // The caller is responsible for setting PR labels computed here.
+    async _doProcess(anotherPrWasStaged) {
+        await this._loadTag();
+        await this._loadLabels();
+        await this._loadPrState();
+        this._log("PR state calculated: " + this._prState.toString());
+        let result = await this.update();
+        if (anotherPrWasStaged)
+            return result.delayed() ? result : StepResult.Fail();
```
rousskov: Why return "fail" if an update was successful?

src/MergeContext.js
```diff
@@ -855,107 +960,172 @@ class MergeFinalizer extends MergeContext {
...
+    // Maintain Anubis-controlled PR metadata.
+    // If possible, also merge or advance the PR towards merging.
+    // The caller is responsible for setting PR labels computed here.
+    async _doProcess(anotherPrWasStaged) {
+        await this._loadTag();
+        await this._loadLabels();
+        await this._loadPrState();
+        this._log("PR state calculated: " + this._prState.toString());
+        let result = await this.update();
+        if (anotherPrWasStaged)
```
rousskov: Should we assert that update() never "suspends" the PR (i.e., never prevents another PR from being merged)?

src/MergeContext.js
```diff
@@ -855,107 +960,172 @@ class MergeFinalizer extends MergeContext {
...
+            return result.delayed() ? result : StepResult.Fail();
+
+        if (this._prState.preStaged()) {
+            result = await this._stage();
+            if (!result.succeeded())
+                return result;
+        }
```
rousskov: If we can assert that the successfully stage()d PR is staged(), then please assert that here.

src/MergeContext.js
```diff
@@ -855,107 +960,172 @@ class MergeFinalizer extends MergeContext {
...
 
-        this._log("merged successfully");
-        return await this._cleanupMerged();
+        if (this._prState.staged()) {
+            result = await this._mergeStaged();
+            if (!result.succeeded())
+                return result;
+        }
```
rousskov: If we can assert that the successfully mergeStage()d PR is postStaged(), then please assert that here.

src/PrMerger.js
```diff
@@ -76,32 +85,6 @@ class PrMerger {
...
-                assert(this.rerunIn);
-                return true;
-            }
-
-            assert(result.succeeded());
-        }
-        return false;
-    }
 
     // Loads 'being-in-merge' PR, if exists (the PR has tag and staging_branch points to the tag).
```
rousskov: I would clarify the intent along these lines:

```
// returns raw PR with a commit at the tip of the staging branch (or null)
// if that PR exists, it is either "staged" or "post-staged"
```

src/MergeContext.js
```diff
@@ -198,19 +200,156 @@ class StatusChecks
...
+            return "no pending labels";
+        return "pending labels: " + str;
+    }
+}
+
+// manage pull request states, one of:
+// pre-staged: prior to staged commit creation
+// staged: prior to staged commit merging into base
+// post-staged: up to all post-merge cleanup completion
+class PrState
```
rousskov: If this is true, please add something like this

```
// staged commit is defined as a not-yet-merged PR commit at the tip of the auto branch
```
```

File: src/MergeContext.js (hunk 13 above)

File: src/MergeContext.js [hunk 17]
```diff
@@ -855,107 +961,176 @@ class MergeFinalizer extends MergeContext {
...
+    // The caller is responsible for setting PR labels computed here.
+    async _doProcess(anotherPrWasStaged) {
+        await this._loadTag();
+        await this._loadLabels();
+        await this._loadPrState();
+        this._log("PR state calculated: " + this._prState.toString());
+        let result = await this.update();
+        if (anotherPrWasStaged) {
+            assert(!result.suspended());
+            return result.delayed() ? result : StepResult.Fail();
```
rousskov: Please always return `result` here. This was discussed in an earlier [review](https://github.com/measurement-factory/anubis/pull/25#discussion_r200009331), and I understand that you were trying to reserve `success` for PRs that were fully merged, but I do not think that translation is used/useful in the current code (AFAICT you agree with that), and it requires extra code (that confuses readers like me and would require a comment to define _doProcess() result).

Let's simplify by always returning result until the caller starts caring about the outcome (other than "suspended PR"). If that ever happens, we would know what the caller new needs are and will adjust this code accordingly.

src/MergeContext.js
```diff
@@ -855,107 +961,176 @@ class MergeFinalizer extends MergeContext {
...
+        if (this._prState.staged()) {
+            result = await this._mergeStaged();
+            if (!result.succeeded())
+                return result;
+            assert(this._prState.postStaged());
+        }
 
+        if (this._prState.postStaged()) {
+            this._role = "finalizer";
+            this._labelMerged();
```
rousskov: This _labelMerged() operation should be a part of _finalize() method because that method is where we process postStaged() PRs. With other changes discussed in this review, it is possible that _finalize() will be called here, just like any other PR processing step, but even if _finalize() has to be called later/specially, I think this _labelMerged() call would belong there.

src/RepoMerger.js
```diff
@@ -96,6 +96,10 @@ class RepoMerger {
 
     _plan(ms) {
         assert(ms > 0);
+        // obey node.js setTimeout() limits
+        const maxMs = Math.pow(2, 31) - 1;
+        ms = ms < maxMs ? ms : maxMs;
```
rousskov:
```suggestion
        const ms = Math.min(requestedMs, maxMs);
```

and rename the parameter to requestedMs, of course.

src/MergeContext.js
```diff
@@ -198,19 +200,156 @@ class StatusChecks
...
+                str += label.toString();
+            }
+        }
+        if (!str.length)
+            return "no pending labels";
+        return "pending labels: " + str;
+    }
+}
+
+// manage pull request states, one of:
```
```suggestion
// A state of a pull request (with regard to merging progress). One of:
```

src/MergeContext.js
```diff
@@ -4,7 +4,9 @@ const Log = require('./Logger.js');
 const GH = require('./GitHubUtil.js');
 const Util = require('./Util.js');
 
-// A result produced by MergeInitiator and MergeFinalizer classes
+// Action outcome, with support for paused and snoozed actions.
+// For example, is returned by PullRequest.process() (at higher level) or
```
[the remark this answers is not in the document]

src/MergeContext.js

src/MergeContext.js
```diff
@@ -377,124 +516,225 @@ class MergeContext {
...
-            await this._removeLabelsIf([label]);
-        else
-            await this._addLabel(label);
+
+        this._log(this._labels.toString());
+        if (this._labels)
+            await this._labels.apply();
+    }
+
+    // Atomize PR post-merge cleanup: suspend this PR
```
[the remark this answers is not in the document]

src/MergeContext.js
```diff
@@ -377,124 +516,225 @@ class MergeContext {
...
+            assert(this._prState.postStaged());
+            await this._applyLabels();
+            await GH.updatePR(this._prNumber(), 'closed');
+            if (this._prState.postStaged())
+                await GH.deleteReference(this._stagingTag());
+            this._log("finalize completed");
+            return StepResult.Succeed();
+        } catch (e) {
+            Log.LogException(e, this._toString() + " unexpected error while finalizing.");
+            return StepResult.Suspend();
```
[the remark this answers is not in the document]

src/MergeContext.js

src/MergeContext.js

src/MergeContext.js
```diff
@@ -855,107 +960,172 @@ class MergeFinalizer extends MergeContext {
...
+
+    async process(anotherPrWasStaged) {
+        let result = null;
+        try {
+            result = await this._doProcess(anotherPrWasStaged);
+        } catch (e) {
+            await this._applyLabels();
+            throw e;
+        }
+        if (!this._prState.postStaged())
```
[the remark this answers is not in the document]

src/RepoMerger.js
eduard-bagdasaryan: Done.

src/MergeContext.js
```diff
@@ -198,19 +200,157 @@ class StatusChecks
...
+        this._prNum = prNum;
+        this._labels = [];
+        for (let label of labels)
+            this._doCache(label.name, true, true);
+    }
+
+    _doCache(name, on, old) {
+        const label = this._labels.find(lbl => lbl.name === name);
+        if (label)
+            label.on = on;
```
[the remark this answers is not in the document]
eduard-bagdasaryan: I left this method, renaming to modify(): it helps to avoid duplication between add() and remove().

src/PrMerger.js
```diff
@@ -54,20 +53,30 @@ class PrMerger {
...
             try {
-                const pr = prList.shift();
+                const rawPr = this._prList.shift();
                 this.total++;
-                const result = await pr.anubisProcessor.process();
-                if (!this.prDone(pr.anubisProcessor, result))
-                    return true;
+                let pr = new PullRequest(rawPr, suspendedEarlier);
+                const result = await pr.process();
+                suspendedEarlier = suspendedEarlier || result.suspended();
```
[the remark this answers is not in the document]
rousskov: Not quite done: If _doProcess() throws, the process() method now _throws_  instead of returning a valid result. I do not think we want process() to throw. At the very least, its previous version caught exceptions instead of throwing IIRC. Note that the "finally" block does not swallow exceptions like the "catch" block does. You probably want both blocks, with the "catch" block returning some reasonable result after logging the exception.

src/MergeContext.js
```diff
@@ -859,107 +962,175 @@ class MergeFinalizer extends MergeContext {
...
+
+        this._unlabelPreconditionsChecking();
+
+        const conditions = await this._checkPreconditions();
+
+        if (!conditions.succeeded()) {
+            if (conditions.delayed())
+                return conditions;
+            // for failed() or suspended()
+            return StepResult.Fail();
```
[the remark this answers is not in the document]
rousskov: > we cannot, since the 'suspended' is used by the higher level code to take the decision whether a PR is being staged

Please fix that (i.e., fix process() caller to use PullRequest::staged() to determine whether the PR is staged) and then come back to this change request. I see no reason to assign a special meaning to the "suspended" result of a process() call. Suspended result should mean that the requested operation could not complete immediately and is waiting for some external non-time event. We should not piggyback "or the PR got staged" on top of that.

src/MergeContext.js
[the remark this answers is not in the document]
rousskov: AFAICT, the _only_ duplication is the complex .find() call (which can and should be converted into a simple this.find() method call. 

Please split and move _modify() logic into the corresponding methods that currently convert meaningful "add" and "remove" names into a rather awkward mustBeSet boolean parameter. It will make the code a lot easier to comprehend.

The only reason to keep _modify() (as a public modify() wrapper that simply calls the other two methods) is to simplify _labelFailedDescription(). Your call whether to do that or not.

src/MergeContext.js
```diff
@@ -377,124 +517,220 @@ class MergeContext {
...
+    async update() {
+        if (this._role !== "initiator")
             return;
-        const label = Config.failedDescriptionLabel();
-        if (isValid)
-            await this._removeLabelsIf([label]);
-        else
-            await this._addLabel(label);
+
+        this._role = "updater";
```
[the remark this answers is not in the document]
rousskov: When you push into _breadcrumbs (in any method), push as early as possible. Doing so will tell you which calls the problematic pull request went through (or did not). In nearly all cases, that push should be the first thing that happens in that method, even before you check various state assertions and before you call update(). This mechanism gives you a history of calls.

We can make an exception for update() -- push only if the method is actually going to run. That is what update() does now. That update() code can stay if you like it.

src/PrMerger.js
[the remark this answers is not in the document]
eduard-bagdasaryan: > If _doProcess() throws, the process() method now throws instead of returning a valid result.

That is what we need: this exception should go into the caller PrMerger.runStep() in order to go ahead with another PR.

>  At the very least, its previous version caught exceptions instead of throwing

No, its previous version (without 'finally') caught the exception, flushed pending labels and rethrew the exception.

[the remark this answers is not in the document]
rousskov: Got it. Sorry for the noise.

eduard-bagdasaryan:
> Please also address the following TODOs and XXXs (the others are out of scope for now):

Done.

src/MergeContext.js
```diff
@@ -4,55 +4,73 @@ const Log = require('./Logger.js');
...
-        return this._delayMs === null;
-    }
+// Relaying process() outcome to the caller via doProcess() exceptions:
+//
+// Exception          Abandon  Push-Labels  Result-of-process()
+// _exLostControl()     yes      no           approval delay (if any)
+// _exObviousFailure()  yes      yes          approval delay (if any)
+// _exLabeledFailure()  yes      yes          approval delay (if any)
+// _exSuspend()         no       yes          approval delay (if any)
+// _exRetry()           yes      yes          retry or approval delay
```
rousskov: Instantly retried PRs are not really abandoned and probably should not flip/flop labels. I suggest:

```suggestion
// _exRetryNow()        no       no           zero delay
// _exRetrySoon()       yes      yes          retry or approval delay
```

And adjust the rest of the code to match the above. For example, _exRetrySoon() will have a hard-coded delay calculation (instead of doing that in the thrower).

src/PrMerger.js
```diff
@@ -5,127 +5,146 @@ const Logger = Log.Logger;
...
+        if (pr.state === 'open')
+            return pr;
+        Logger.warn("The current PR" + prNum + " unexpectedly closed.");
+        return null;
     }
 } // PrMerger
 
-module.exports = PrMerger;
+// PrMerger's create and execute() wrapper.
+module.exports = async function Step() {
```
rousskov: If possible, please split into two statements. The Step() function definition is complex enough on its own to keep it isolated from the exporting concerns/statement.

src/MergeContext.js
```diff
@@ -859,107 +1015,302 @@ class MergeFinalizer extends MergeContext {
...
-            return StepResult.Succeed();
         } catch (e) {
             if (e.name === 'ErrorContext' && e.unprocessable()) {
-                if (await this._tagDiverged()) {
-                    Log.LogException(e, this._toString() + " fast-forwarding failed");
-                    return await this._cleanupMergeFailed(true);
-                }
+                await this._stagedPosition().compute();
+                if (this._stagedPosition.diverged())
+                    this._log("could not fast-forward, the base " + this._prBaseBranchPath() + " seems to be modified just now");
```
```suggestion
                    this._log("could not fast-forward, the base " + this._prBaseBranchPath() + " was probably modified while we were merging");
```

src/MergeContext.js
```diff
@@ -198,19 +216,269 @@ class StatusChecks
...
+    merged() {
+        assert(this._status);
+        return this._status === "identical" || this._status === "behind";
+    }
+
+    // neither ahead nor merged:
+    // both base and the feature branch have unique-to-them commits
+    diverged() {
+        assert(this._status);
+        return this._status === "diverged";
```
rousskov: To better match our method description and to be on the safe side in case GitHub adds more statuses:

```suggestion
         // this._status should be "diverged", but we also handle any new/unexpected statuses here
        return !(this.ahead() || this.merged());
```

N.B. I am curious, what status will GitHub report if the commits have no common ancestor at all? I understand that this is highly unlikely in a mature repository, but I think it is technically possible if somebody force-pushes a completely different/new master branch after the feature branch was branched off the old master. Would that result in a "diverged" `_status`? The change suggested above would make the answer irrelevant.

src/MergeContext.js
```diff
@@ -570,193 +882,61 @@ class MergeContext {
...
         }
 
-        if (this._approval.grantedTimeout())
-            return StepResult.Delay(this._approval.delayMs);
-
-        if (!statusChecks.final())
-            return StepResult.Suspend();
-
-        return StepResult.Succeed();
+        assert(!this._restrictions.stagingBanned());
```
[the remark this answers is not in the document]

src/MergeContext.js

src/PrMerger.js
eduard-bagdasaryan: Done.

src/MergeContext.js
```diff
@@ -364,137 +634,161 @@ class MergeContext {
         return this._tagCommitCache;
     }
 
-    // Whether the PR merge commit has not changed since the PR staged commit creation.
+    // Calculates whether the PR staged commit is in sync with the PR branch.
```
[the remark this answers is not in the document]
eduard-bagdasaryan: I also described that the result also depends on the staged commit position (related to the base).

src/MergeContext.js
```diff
@@ -364,137 +634,161 @@ class MergeContext {
...
+    // (no new branch commits since PR staged commit creation).
```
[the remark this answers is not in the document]

src/MergeContext.js
```diff
@@ -364,137 +634,161 @@ class MergeContext {
...
-        const result = tagCommit.tree.sha === prCommit.tree.sha;
-        this._log("tag freshness: " + result);
-        return result;
+        const tagInSyncWithPrBranch = tagCommit.tree.sha === prCommit.tree.sha;
+        this._log("Merge commit and PR branch synchronization: " + tagInSyncWithPrBranch);
+        return tagInSyncWithPrBranch;
     }
 
-    // Label manipulation methods
+    // loads info about "staging tag" (if any); M-staged-PRnnn tag is a
```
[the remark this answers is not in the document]

src/MergeContext.js
```diff
@@ -570,193 +882,61 @@ class MergeContext {
...
+            return;
         }
 
-        if (!this._approval.granted()) {
-            this._logFailedCondition("approved");
-            return StepResult.Fail();
+        if (!this._stagingSha)
+            this._stagingSha = await GH.getReference(Config.stagingBranchPath());
+        // Make sure the staging branch points to our fresh staged commit.
+        if (this._stagingSha !== this._tagSha) {
```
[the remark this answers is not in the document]

src/MergeContext.js
```diff
@@ -1243,11 +1242,21 @@ class PullRequest {
 
     /* _ex*() methods below are mutually exclusive: first match wins */
 
+    // instant retries does not require reprocessing from scratch
+    _exRetryNow(why) {
+        assert(arguments.length === 1);
+        this._reprocessingDelayMs = 0;
```
rousskov: Delay value aside, _exRetryNow() looks just like _exRetrySoon(), but they are supposed to have very different effect. See the `Push-Labels` column values in the exceptions summary table. AFAICT, that column is correct (and has different values for these two exceptions).

src/MergeContext.js
```diff
@@ -840,7 +842,7 @@ class PullRequest {
     staged() { return this._prState.staged(); }
 
     _debugString() {
-        const staged = this._stagingSha ? "staged: " + this._stagingSha.substr(0, this._shaLimit) + ' ' : "";
+        const staged = this._tagSha ? "staged: " + this._tagSha.substr(0, this._shaLimit) + ' ' : "";
```
rousskov: Should the `staged` string be limited to _staged_ PRs?

* If yes:
```suggestion
        const staged = this.staged() ? "staged: " + this._tagSha.substr(0, this._shaLimit) + ' ' : "";
```
* Otherwise:
```suggestion
        const tagged = this._tagSha ? "tagged: " + this._tagSha.substr(0, this._shaLimit) + ' ' : "";
```

eduard-bagdasaryan: > Should the `staged` string be limited to _staged_ PRs?

I would print this tag sha in any case, which may be useful. I renamed to 'tagged' as you suggested.

src/MergeContext.js
eduard-bagdasaryan: Agreed, I forgot to set _labelPushBan. Done.

src/MergeContext.js
```diff
@@ -198,19 +217,269 @@ class StatusChecks
...
+        await GH.addLabels(params);
+    }
+
+    _find(name) { return this._labels.find(label => label.name === name); }
+}
+
+// A state of an open pull request (with regard to merging progress). One of:
+// brewing:
+//   a) without a staged commit (PRs are created in this state);
+//   b) with a fresh staged commit having failed checks, which is kept until it becomes stale
```
rousskov: Do not mix definitions with consequences/actions based on those definitions:

```suggestion
//   b) with a fresh staged commit having failed checks
```

src/MergeContext.js
```diff
@@ -198,19 +217,269 @@ class StatusChecks
...
+//   c) with a stale staged commit which will be deleted before creation a new one
```
```suggestion
//   c) with a stale staged commit
```

src/MergeContext.js
```diff
@@ -364,137 +599,167 @@ class MergeContext {
...
+        if (!this._approval.granted())
+            throw this._exSuspend("waiting for approval");
+
+        if (this._approval.grantedTimeout())
+            throw this._exSuspend("waiting for objections");
+
+        if (this._stagingBanned)
+            throw this._exSuspend("waiting for another staged PR");
+
+        if (this._stagedCommitWillFail)
```
rousskov: It feels like this check should go above the _approval.granted() check because, without a label, humans are likely to miss this poorly visible failure and may end up waiting for votes and other PRs instead of fixing the broken PR branch.

src/MergeContext.js
```diff
@@ -1126,10 +1126,6 @@ class PullRequest {
     async _doProcess() {
         this._breadcrumbs.push("load");
 
-        /*
-         * Until _loadRawPr(), we must avoid this._rawPr fields except .number.
```
rousskov: I do not think this problem has been addressed in 38574c: It is still possible to access stale `this._rawPr` fields before _loadRawPr() loads fresh PR metadata. I suggest restoring that TODO. I do not recommend implementing that TODO during this project though.

src/PrMerger.js
```diff
@@ -145,8 +145,9 @@ class PrMerger {
 // PrMerger's create and execute() wrapper.
 async function Step() {
     let mergerer = new PrMerger();
-    return mergerer.execute();
-};
+    const result = await mergerer.execute();
```
rousskov: I do not think this `await` was missing -- the old code returned a promise, which any async call is allowed to do. What is probably not needed here is the `async` designation -- this function is not an async function. I suggest restoring the old no-`await` code, removing `async`, and changing the Step() description to something like

```
// promises to process all PRs once, hiding PrMerger from callers
```

src/MergeContext.js
```diff
@@ -859,107 +987,268 @@ class MergeFinalizer extends MergeContext {
...
 
-        this._log("merged successfully");
-        return await this._cleanupMerged();
+            let result = new ProcessResult();
 
+            if (this._prState.staged()) {
+                if (suspended)
+                    result.setPrStaged(true);
+                else
+                    this._removeStagingLabels(); // abandoning this PR
```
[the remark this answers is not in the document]

src/MergeContext.js
```diff
@@ -859,107 +987,268 @@ class MergeFinalizer extends MergeContext {
...
+            }
+
+            if (knownProblem) {
+                result.setDelayMsIfAny(this._delayMs()); // may be null
```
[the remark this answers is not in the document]

src/MergeContext.js
```diff
@@ -859,107 +987,268 @@ class MergeFinalizer extends MergeContext {
...
+    // reprocessing will be required after the problem is fixed
+    _exLabeledFailure(why, label) {
+        assert(arguments.length === 2);
+        assert(!this._labelPushBan);
+        this._labels.add(label);
+        return new PrProblem(why);
+    }
+}
+
+// PullRequest::process() wrapper
```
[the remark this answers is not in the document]
eduard-bagdasaryan: Done differently, see 9e982a.

src/MergeContext.js
rousskov: > If this TODO's scope is only doProcess() method

No, the problem is not confined to doProcess(). The doProcess() location makes sense IMO because it is there we transition from dangerous inconsistent state to a consistent one, but it is possible that there is a better location for that TODO -- my change request did not consider the location.

> It is still possible, of course, to insert there a dozen of other methods, requiring some other rawPr fields and thus break everything.

Yes, and, somewhat more likely, accidentally access stale rawPr fields when adding code to the _existing_ methods. IIRC, one of your recent commits did exactly that.

> However, following this logic, we would need to add such 'TODOS' in many other places where similar data dependencies take place.

One TODO per object is usually enough. I hope that rawPr is the only stale object that the code may accidentally access, but please correct me if I am wrong. If other such objects exist, then it may be indeed appropriate to flag them with XXXs or TODOs, especially if they stay stale across methods.

src/MergeContext.js
```diff
@@ -859,107 +990,271 @@ class MergeFinalizer extends MergeContext {
...
+    // reprocessing will be required after the problem is fixed
+    _exLabeledFailure(why, label) {
+        assert(arguments.length === 2);
+        assert(!this._labelPushBan);
+        this._labels.add(label);
+        return new PrProblem(why);
+    }
+}
+
+// promises to process a single PR at once, hiding PullRequest from callers
```
```suggestion
// promises to update/advance the given PR, hiding PullRequest from callers
```

src/MergeContext.js
```diff
@@ -570,248 +856,89 @@ class MergeContext {
...
+        if (this._stagedPosition.merged()) {
+            this._log("already merged into base some time ago");
+            this._prState = PrState.Merged();
+            return;
         }
 
-        if (!this._prMergeable()) {
-            this._logFailedCondition("mergeable");
-            return StepResult.Fail();
+        // The staged commit became out of sync with PR and(or) base branches.
```
rousskov: Please remove this comment -- the log message below it should speak for itself.

src/MergeContext.js
```diff
@@ -570,248 +856,89 @@ class MergeContext {
...
+        if (!(await this._stagedCommitIsFresh())) {
+            this._log("the staged commit became stale due to PR branch and(or) base branch changes");
```
rousskov: This log message became stale. I do not think we should enumerate all _stagedCommitIsFresh() checks here:

```suggestion
            this._log("the staged commit is stale");
```

src/MergeContext.js
```diff
@@ -198,19 +217,233 @@ class StatusChecks
...
+
+        // major methods we have called, in the call order (for debugging only)
+        this._breadcrumbs = [];
+
+        this._messageValid = null;
+
+        this._prState = null; // calculated PrState
+
+        this._labels = null;
+        // while unexpected, PR merging and closing is not prohibited when staging is
```
rousskov: To clarify the scope of this comment, please add new lines:
```suggestion

        // while unexpected, PR merging and closing is not prohibited when staging is
        this._stagingBanned = banStaging;
        
```

src/MergeContext.js
rousskov: Nice!

src/MergeContext.js
```diff
@@ -364,137 +599,170 @@ class MergeContext {
...
 
-    // Whether the PR merge commit has not changed since the PR staged commit creation.
-    // Note that it does not track possible conflicts between PR base branch and the
-    // PR branch (the PR merge commit is recreated only when there are no conflicts).
-    // Conflicts are tracked separately, by checking _prMergeable() flag.
-    async _tagIsFresh() {
+    // Determines whether the existing staged commit is equivalent to a staged
+    // commit that could be created right now. Relies on PR merge commit being fresh.
+    // TODO: lacks some checks, e.g., mismatching staging checks list (configured
+    // on GitHub) or staging commit message (Anubis now may generate a different message).
```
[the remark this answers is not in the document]
eduard-bagdasaryan: I misinformed you with this part of TODO: in reality the message generation check already exists by means of _messageIsFresh().

src/MergeContext.js

src/MergeContext.js

src/MergeContext.js

src/MergeContext.js
```diff
@@ -364,137 +599,170 @@ class MergeContext {
...
     }
 
-    // Whether the PR merge commit has not changed since the PR staged commit creation.
-    // Note that it does not track possible conflicts between PR base branch and the
-    // PR branch (the PR merge commit is recreated only when there are no conflicts).
-    // Conflicts are tracked separately, by checking _prMergeable() flag.
-    async _tagIsFresh() {
+    // Determines whether the existing staged commit is equivalent to a staged
+    // commit that could be created right now. Relies on PR merge commit being fresh.
+    // TODO: lacks some checks, e.g., mismatching staging checks list (configured
```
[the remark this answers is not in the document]
rousskov: Thank you for explaining this so well. Resolved.

src/MergeContext.js
```diff
@@ -541,24 +556,27 @@ class PullRequest {
...
         if (!Config.manageApprovalStatus())
             return;
 
-        const combinedStatus = await GH.getStatuses(sha);
-        const approvalStatus = combinedStatus.statuses ?
-            combinedStatus.statuses.find(el => el.context.trim() === Config.approvalContext()) : null;
-
-        if (approvalStatus && this._approval.matchesGitHubStatusCheck(approvalStatus)) {
-            this._log("approval status already exists: " + Config.approvalContext() + ", " + this._approval);
+        if (this._dryRun("setting approval status"))
```
rousskov: It looks like this check should be moved into _createApprovalStatus() because hasApprovalStatus() and addApprovalStatus() calls do not modify GitHub (and do allow us to get deeper in dry-run mode).

PR #27, opened 2018-09-03
Properly handle PRs with empty descriptions

eduard-bagdasaryan:
Before this fix, the bot misinterpreted null message bodies.

Also:
* correctly compare commit messages with empty descriptions
* ping event logging (this event is sent during webhook creation)

### DIFF
--- a/src/MergeContext.js
+++ b/src/MergeContext.js
@@ -494,7 +494,7 @@ class MergeContext {
     _prHeadSha() { return this._pr.head.sha; }
 
     _prMessage() {
-        return this._pr.title + ' (#' + this._pr.number + ')' + '\n\n' + this._prBody();
+        return (this._pr.title + ' (#' + this._pr.number + ')' + '\n\n' + this._prBody()).trim();
     }
 
     _prMessageValid() {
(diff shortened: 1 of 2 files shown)


src/MergeContext.js
```diff
@@ -494,7 +494,7 @@ class MergeContext {
     _prHeadSha() { return this._pr.head.sha; }
 
     _prMessage() {
-        return this._pr.title + ' (#' + this._pr.number + ')' + '\n\n' + this._prBody();
+        return (this._pr.title + ' (#' + this._pr.number + ')' + '\n\n' + this._prBody()).trim();
```
rousskov: I think the intent here is not clear. I would replace .trim() with message updates:

```
let message = ...
if (this._prBody().length)
    message += "\n\n" + this._prBody();
return message;
```

but I do not insist on this or a similar change.

src/MergeContext.js
```diff
@@ -527,7 +527,7 @@ class MergeContext {
 
     _prOpen() { return this._pr.state === 'open'; }
 
-    _prBody() { return this._pr.body.replace(/\r+\n/g, '\n'); }
+    _prBody() { return this._pr.body ? this._pr.body.replace(/\r+\n/g, '\n') : ""; }
```
rousskov: I would worry that some non-null pr.body values/text might be [falthy](https://developer.mozilla.org/en-US/docs/Glossary/Falsy). Please use an explicit comparison with null (and, if needed, undefined).

src/MergeContext.js
eduard-bagdasaryan: Left as is because the suggested approach would not work when this._prBody() contains only whitespaces (its length > 0).

src/MergeContext.js

src/MergeContext.js
rousskov: _prBody() should return trimmed bodies, of course.

PR #28, opened 2018-10-24
Properly handle getReference() API call results

eduard-bagdasaryan:
Before this fix, Anubis assumed that getReference() would return the
result only on exact ref matching. This is is not true: if the ref doesn't
exist in the repository, but there are other refs starting with the
ref, they are returned as an array. This resulted in an unhandled
exception within the promise and getting stuck on unresolved promise. We
should explicitly reject such mis-results.

Also I added a process unhandledRejection event handler to make it
easier to detect such bugs in the future.

### DIFF
--- a/src/GitHubUtil.js
+++ b/src/GitHubUtil.js
@@ -219,6 +219,12 @@ function getReference(ref) {
                 reject(new ErrorContext(err, getReference.name, params));
                 return;
             }
+            // If the ref does not exist, but  but existing refs start with ref,
+            // an array is returned.
+            if (res.data === undefined || res.data.object === undefined) {
+                reject(new ErrorContext("Could not find " + params.ref + " reference", getReference.name, params));
+                return;
+            }
             const result = {ref: res.data.ref, sha: res.data.object.sha};
             logApiResult(getReference.name, params, result);
             resolve(res.data.object.sha);
--- a/src/GitHubUtil.js
+++ b/src/GitHubUtil.js
@@ -219,9 +219,9 @@ function getReference(ref) {
                 reject(new ErrorContext(err, getReference.name, params));
                 return;
             }
-            // If the ref does not exist, but  but existing refs start with ref,
-            // an array is returned.
-            if (res.data === undefined || res.data.object === undefined) {
+            // If the requested ref does not exist in the repository, but some
+            // existing refs start with it, they will be returned as an array.
+            if (Array.isArray(res.data)) {
                 reject(new ErrorContext("Could not find " + params.ref + " reference", getReference.name, params));
                 return;
             }
--- a/src/Main.js
+++ b/src/Main.js
@@ -9,6 +9,7 @@ const WebhookHandler = createHandler({ path: Config.githubWebhookPath(), secret:
 
 process.on('unhandledRejection', error => {
     Logger.error("unhandledRejection", error.message);
+    process.exit(1);
 });
 
 // events
--- a/src/Main.js
+++ b/src/Main.js
@@ -7,6 +7,10 @@ const Logger = Log.Logger;
 
 const WebhookHandler = createHandler({ path: Config.githubWebhookPath(), secret: Config.githubWebhookSecret() });
 
+process.on('unhandledRejection', error => {
+    Logger.error("unhandledRejection", error.message);
+});
+
 // events
 
 WebhookHandler.on('error', (err) => {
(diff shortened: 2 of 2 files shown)


src/Main.js
```diff
@@ -7,6 +7,10 @@ const Logger = Log.Logger;
 
 const WebhookHandler = createHandler({ path: Config.githubWebhookPath(), secret: Config.githubWebhookSecret() });
 
+process.on('unhandledRejection', error => {
+    Logger.error("unhandledRejection", error.message);
```
rousskov: I think we should quit/abort the whole Anubis process on these errors. Would that be difficult?

src/GitHubUtil.js
```diff
@@ -219,6 +219,12 @@ function getReference(ref) {
                 reject(new ErrorContext(err, getReference.name, params));
                 return;
             }
+            // If the ref does not exist, but  but existing refs start with ref,
+            // an array is returned.
+            if (res.data === undefined || res.data.object === undefined) {
```
rousskov: If possible, please use some direct test like `res instanceof Array`.

Otherwise, replace `foo === undefined` with `!foo` to simplify and capture more unexpected cases.

eduard-bagdasaryan: Done with Array.isArray() check.

src/Main.js
eduard-bagdasaryan: Done.

rousskov: As discussed during the meeting, please check whether process.exit() is the best way to immediately stop the program execution.

eduard-bagdasaryan: The [documentation](https://nodejs.org/api/process.html#process_process_exit_code) says that it "forces the process to exit as quickly as possible even if there are still asynchronous operations pending that have not yet completed fully". I do not see any problems with it. Quick googling did not show such potential problems either.

eduard-bagdasaryan: In my tests I have not noticed any differences: in both cases, the program terminates and logs are saved. However, if the docs say that rethrowing is probably more reliable (that logs may be dropped sometimes due to process.exit()), we should use the rethrow approach. I found another related [discussion](https://stackoverflow.com/questions/5266152/how-to-exit-in-node-js), where some people recommend using this approach either.
I pushed that change at 0141804.

eduard-bagdasaryan: Another case, which is uncaughtException should be covered also. In my tests, I see problems with logging in this handler: the logs are printed into stderr/stdout, but the log file still lacks them There is a general problem, discussed [here](https://github.com/trentm/node-bunyan/issues/37). The possible solution, using [exeunt](https://github.com/joyent/node-exeunt) module, works only for stderr/stdout, and does not work for files.

rousskov: I do not think we should risk post-exception execution of event handlers or spend time on trying to prevent that all such executions except for logging event handlers. It feels like we would be solving an unsolvable problem.

Instead, to retain crash information, I suggest that we use [node --abort-on-unhandled-exception](https://nodejs.org/dist/v8.9.3/docs/api/cli.html#cli_abort_on_uncaught_exception) approach as [mentioned](https://github.com/trentm/node-bunyan/issues/37#issuecomment-298810505) on the Bunyan thread you found. AFAICT, that approach requires node v0.10. Are we already using that (or later) node version? If not, we can wait until we upgrade.

If `--abort-on-unhandled-exception` does not prevent us from using a custom uncaught exception handler, then, in _addition_ to using that node option, please install that handler to log a brief, safe message about dying on an uncaught exception to stderr, even though that message is likely to be lost when Anubis is started from a system script (as it should be).

eduard-bagdasaryan: > Are we already using that (or later) node version?

That is a rather old version, we use now v9.0.0 or later.

> If --abort-on-unhandled-exception does not prevent us from using a custom uncaught exception handler,
With this option, the custom `uncaughtException` handler is not called

I restarted the bot with `--abort-on-unhandled-exception` on the server, and now we can only notice a crash by core files generated. How should we limit core file size there?

rousskov: > we can only notice a crash by core files generated

Is it fairly easy to get usable crash information (e.g., which exception was not caught and where it was thrown) from the core file?

> How should we limit core file size there?

I am not a sysadmin, and I do not know about that specific server policies. Can we just use whatever setting are already configured there? Or are the core dumps disabled on that server right now? At any rate, please see if you and kinkie (who, I presume, administers that server) can agree on a reasonable setting, whatever it is (e.g., unlimited size with the last few preserved).

eduard-bagdasaryan: > Is it fairly easy to get usable crash information

I found an [article](https://www.javascriptjanuary.com/blog/nodejs-postmortem-debugging-for-fun-and-production) describing such debugging with lldb + llnode. Looks not hard, but I have not experimented with it (which can be done when required).

> Can we just use whatever setting are already configured there?

I tried to force the bot to generate core file and it worked there. So it looks like no ulimit adjustments needed there.

PR #30, opened 2019-03-18
Removed PR tags from the implementation

eduard-bagdasaryan:
We could not use PR tags anymore because new Git versions report errors
when encountering tag modifications.

Basically, PR tags helped with two problems:

* distinguish merged (but still open) PRs, in order to avoid processing
  this PR from scratch. Instead of this, we now track such PRs by
  searching the base branch commits for the ' (#NNN)' commit title
  suffix.

* mark failed staged commits and skip them (no modifications), thus
  avoiding the endless staged commit recreation. We now use
  M-failed-staged-checks label for this purpose: the bot ignores PRs
  marked with this label until a human clears it off; the bot itself
  does not clear it.

### DIFF
--- a/src/MergeContext.js
+++ b/src/MergeContext.js
@@ -672,7 +672,6 @@ class PullRequest {
         this._log("staged commit does not exist");
     }
 
-
     async _loadLabels() {
         let labels = await GH.getLabels(this._prNumber());
         assert(!this._labels);
@@ -897,8 +896,10 @@ class PullRequest {
         let commits = null;
 
         if (this._prBaseBranch() === 'master') {
-            const query = 'repo:' + Config.owner() + "/" + Config.repo() + '+#' + this._prNumber() +
-                '+author:' + this._prAuthor() + '+committer-date:>' + dateSince;
+            const query = 'repo:' + Config.owner() + "/" + Config.repo() +
+                '+ (#' + this._prNumber() + ')' +
+                '+author:' + this._prAuthor() +
+                '+committer-date:>' + dateSince;
             commits = await GH.searchCommits(query); // searches the default branch
         } else {
             commits = await GH.getCommits(this._prBaseBranch(), dateSince, this._prAuthor()); // searches any branch
--- a/src/MergeContext.js
+++ b/src/MergeContext.js
@@ -448,7 +448,6 @@ class PullRequest {
         this._requiredContextsCache = null;
 
         this._stagedPosition = null;
-        this._freshStagedCommitWithFailedChecks = null;
 
         // this PR staged commit received from GitHub, if any
         this._stagedCommit = null;
@@ -952,7 +951,7 @@ class PullRequest {
         // Do not vainly recreate staged commit which will definitely fail again,
         // since the PR+base code is yet unchanged and the existing errors still persist
         if (stagedStatuses.failed()) {
-            this._freshStagedCommitWithFailedChecks = true;
+            this._labels.add(Config.failedStagingChecksLabel());
             await this._enterBrewing();
             return;
         }
@@ -1215,9 +1214,9 @@ class PullRequest {
          */
         await this._loadStaged();
         await this._loadRawPr(); // requires this._loadStaged()
-        await this._loadPrState(); // requires this._loadRawPr()
-        this._log("PR state: " + this._prState);
         await this._loadLabels();
+        await this._loadPrState(); // requires this._loadRawPr() and this._loadLabels()
+        this._log("PR state: " + this._prState);
 
         if (this._prState.brewing()) {
             await this._stage();
--- a/src/MergeContext.js
+++ b/src/MergeContext.js
@@ -664,7 +664,7 @@ class PullRequest {
            return;
         const stagedSha = await GH.getReference(Config.stagingBranchPath());
         this._stagedCommit = await GH.getCommit(stagedSha);
-        const prNum = Util.ParsePrMessage(this._stagedCommit.message);
+        const prNum = Util.ParsePrNumber(this._stagedCommit.message);
         if (prNum !== null && this._prNumber().toString() === prNum) {
             this._log("found staged commit " + stagedSha);
             this._stagedSha = stagedSha;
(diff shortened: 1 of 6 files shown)


src/MergeContext.js
```diff
@@ -891,14 +876,53 @@ class PullRequest {
...
+        let d = new Date();
+        d.setDate(d.getDate() - days);
+        const dd = String(d.getDate()).padStart(2, '0');
+        const mm = String(d.getMonth() + 1).padStart(2, '0'); //January is 0!
+        const yyyy = d.getFullYear();
+        return yyyy + '-' + mm + '-' + dd;
+    }
+
+    /// Checks whether the PR base branch has this PR's staged commit merged.
+    /// TODO: for now the search scope of this method is limited only by
```
rousskov: Since Anubis is used for merging non-master PRs, this TODO sounds like an XXX that we have to address before this change goes in.

src/MergeContext.js
```diff
@@ -891,14 +876,53 @@ class PullRequest {
...
+    /// the project 'default branch', configured on GitHub, which is usually
+    /// (but not always) 'master' branch.
+    async _mergedSomeTimeAgo() {
+        const query = 'repo:' + Config.owner() + "/" + Config.repo() + '+#' + this._prNumber() +
```
rousskov: Can we add parenthesis and a space for the full ` (#NNN)` pattern? If not, we should probably add a comment why not.

rousskov: Please reformat to place/add each query parameter on its own line.

src/Util.js
```diff
@@ -13,22 +13,15 @@ function commonParams() {
     };
 }
 
-const TagRegex = /(refs\/)(tags\/M-staged-PR)(\d+)$/;
+const PrNumberRegex = /(.* \(#)(\d+)(\)$)/;
```
rousskov: If we do not need `.*`, please drop that part of the regex.

rousskov: We certainly need to capture the PR number (i.e. the second set of unescaped parenthesis), but it is not clear why we need to capture the prefix string (i.e. why we need the first set of unescaped parenthesis). Remove as unneeded?

Same for the last set of unescaped parenthesis that capture the closing paren and the end-of-line.

src/PrMerger.js
```diff
@@ -93,51 +90,25 @@ class PrMerger {
...
-        // and parse out PR number from the tag name
-        const tag = this._tags.find((t) => { return (t.object.sha === stagingSha) && Util.MatchTag(t.ref); });
-        if (tag === undefined) {
-            Logger.info("No current PR found.");
-            return null;
+        const stagedBranchSha = await GH.getReference(Config.stagingBranchPath());
+        const stagedBranchCommit = await GH.getCommit(stagedBranchSha);
+        Logger.info("Staged branch head sha: " + stagedBranchCommit.sha);
+        const prNum = Util.ParsePrNumber(stagedBranchCommit.message);
+        if (prNum === null) {
```
rousskov: The checks for ParsePrNumber() results are inconsistent: Some callers compare with `null` and some calculate an implicit thruthy-ness of the value. I think it is better _not_ to compare with null explicitly because it will cover the (currently impossible?) case where prNum is undefined. I would also assert that a found prNum is always positive inside ParsePrNumber().

src/MergeContext.js
```diff
@@ -667,23 +658,21 @@ class PullRequest {
...
+        const prNum = Util.ParsePrNumber(stagedCommit.message);
+        if (prNum !== null && this._prNumber().toString() === prNum) {
+            this._log("found staged commit " + stagedSha);
+            this._stagedCommit = stagedCommit;
+            return;
+        }
+        this._log("staged commit does not exist");
     }
 
+
```
rousskov: If we usually use a single line to delimit functions, please remove this extra line.

src/MergeContext.js

src/MergeContext.js

src/PrMerger.js
eduard-bagdasaryan: I reworked this code after addressing another request.

>  I would also assert that a found prNum is always positive inside ParsePrNumber().

Done.

src/Util.js
eduard-bagdasaryan: Done.

eduard-bagdasaryan: Done.

src/MergeContext.js
```diff
@@ -891,14 +875,58 @@ class PullRequest {
...
+        return yyyy + '-' + mm + '-' + dd;
+    }
+
+
+    /// Checks whether the PR base branch has this PR's staged commit merged.
+    async _mergedSomeTimeAgo() {
+        const dateSince = this._dateForDaysAgo(100);
+        let commits = null;
+
+        if (this._prBaseBranch() === 'master') {
```
rousskov: Please add a comment:

```
// Optimization: GitHub can search the master branch by commit message substring
```

src/Util.js
```diff
@@ -13,22 +13,16 @@ function commonParams() {
...
+    const matched = lines[0].match(PrNumberRegex);
     if (!matched)
         return null;
-    return matched[3];
-}
-
-function StagingTag(prNum) {
-    assert(prNum);
-    return "tags/M-staged-PR" + prNum;
+    assert(matched[1] > 0);
```
rousskov: Let's name `matched[1]` prNumber or something like that.

src/MergeContext.js
```diff
@@ -891,14 +875,58 @@ class PullRequest {
...
+    _dateForDaysAgo(days) {
+        let d = new Date();
+        d.setDate(d.getDate() - days);
+        const dd = String(d.getDate()).padStart(2, '0');
+        const mm = String(d.getMonth() + 1).padStart(2, '0'); //January is 0!
+        const yyyy = d.getFullYear();
+        return yyyy + '-' + mm + '-' + dd;
+    }
+
+
```
rousskov: Extra empty line.

src/MergeContext.js
```diff
@@ -368,8 +368,7 @@ class Labels
 //         immediately or after successful tests completion;
 // merged: with a staged commit that has been merged into the base branch.
 // Here, PR "staged commit" is a commit at the tip of the staging branch
-// pointed to by the PR tag. If a PR tag does not exist or does not point to
-// the tip of the staging branch, then the PR does not have a staged commit.
+// with commit message first line, having the PR number NNN (as "(#PRNNN)" string).
```
rousskov: If we require a leading space, then let's mention that here:

```suggestion
// with the commit title ending with the PR number (as " (#PRNNN)" string).
```

but it may be better to refer to the actual regex:


```suggestion
// with commit title ending with the PR number (see PrNumberRegex).
```

src/MergeContext.js

src/Util.js
eduard-bagdasaryan: Done.

src/MergeContext.js

src/MergeContext.js

PR #31, opened 2019-04-03
Prohibit non-ASCII_printable symbols in commit messages

eduard-bagdasaryan:
isprint(3) defines "ASCII printable" symbol range as [0x20, 0x7E].

### DIFF
--- a/src/MergeContext.js
+++ b/src/MergeContext.js
@@ -790,7 +790,7 @@ class PullRequest {
     // Searches for the first non_ASCII_printable character and returns
     // its position in the string.
     _invalidCharacterPosition(str) {
-        const prohibitedCharacters = /[^\x20-\x7e]/; // allow non-special ASCII characters
+        const prohibitedCharacters = /[^\u{20}-\u{7f}]/u; // allow non-special ASCII characters
         const match = prohibitedCharacters.exec(str);
         return match ? match.index : -1;
     }
@@ -799,7 +799,7 @@ class PullRequest {
         const lines = this._prMessage().split('\n');
         for (let i = 0; i < lines.length; ++i) {
             const line = lines[i];
-            // this._prBody() already cared about removing trailing \r characters
+            // this._prBody() already removed trailing \r characters
             const invalidPosition = this._invalidCharacterPosition(line);
             if (invalidPosition !== -1) {
                 this._warn(`PR message has an invalid character at line ${i}, offset ${invalidPosition}`);
--- a/src/MergeContext.js
+++ b/src/MergeContext.js
@@ -789,7 +789,7 @@ class PullRequest {
 
     // returns the position of the first non-ASCII_printable character (or -1)
     _invalidCharacterPosition(str) {
-        const prohibitedCharacters = /[^\u{20}-\u{7e}]/u; // allow non-special ASCII characters
+        const prohibitedCharacters = /[^\u{20}-\u{7e}]/u;
         const match = prohibitedCharacters.exec(str);
         return match ? match.index : -1;
     }
--- a/src/MergeContext.js
+++ b/src/MergeContext.js
@@ -787,6 +787,9 @@ class PullRequest {
         return (this._rawPr.title + ' (#' + this._rawPr.number + ')' + '\n\n' + this._prBody()).trim();
     }
 
+    // Searches for the first non_ASCII_printable character and returns
+    // its position in the string. By default, JavaScript strings are
+    // represented as a sequence of UTF-16 characters.
     _invalidCharacterPosition(str) {
         const prohibitedCharacters = /[^\u0020-\u007f]+/; // allow non-special ASCII characters
         const match = prohibitedCharacters.exec(str);
@@ -799,7 +802,7 @@ class PullRequest {
             const line = lines[i];
             const invalidPosition = this._invalidCharacterPosition(line);
             if (invalidPosition !== -1) {
-                this._warn("PR message has an invalid character at (" + i + "," + invalidPosition + ")");
+                this._warn(`PR message has an invalid character at line ${i}, offset ${invalidPosition}`);
                 return false;
             }
             if (line.length > 72)
--- a/src/MergeContext.js
+++ b/src/MergeContext.js
@@ -788,10 +788,9 @@ class PullRequest {
     }
 
     // Searches for the first non_ASCII_printable character and returns
-    // its position in the string. By default, JavaScript strings are
-    // represented as a sequence of UTF-16 characters.
+    // its position in the string.
     _invalidCharacterPosition(str) {
-        const prohibitedCharacters = /[^\u0020-\u007f]+/; // allow non-special ASCII characters
+        const prohibitedCharacters = /[^\x20-\x7e]/; // allow non-special ASCII characters
         const match = prohibitedCharacters.exec(str);
         return match ? match.index : -1;
     }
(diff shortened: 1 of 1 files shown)


src/MergeContext.js
```diff
@@ -787,9 +787,21 @@ class PullRequest {
         return (this._rawPr.title + ' (#' + this._rawPr.number + ')' + '\n\n' + this._prBody()).trim();
     }
 
+    _invalidCharacterPosition(str) {
```
rousskov: Please document whether this function returns the number of octets/bytes or the number of utf-8 characters leading to the first invalid character.

src/MergeContext.js
```diff
@@ -787,9 +787,21 @@ class PullRequest {
...
+    }
+
     _prMessageValid() {
         const lines = this._prMessage().split('\n');
-        for (let line of lines) {
+        for (let i = 0; i < lines.length; ++i) {
+            const line = lines[i];
+            const invalidPosition = this._invalidCharacterPosition(line);
+            if (invalidPosition !== -1) {
+                this._warn("PR message has an invalid character at (" + i + "," + invalidPosition + ")");
```
```suggestion
                this._warn(`PR message has an invalid character at line ${i}, offset ${invalidPosition}`);
```

eduard-bagdasaryan:
> What is this._prMessage encoding? 

According to the ECMAScript specification, [Section 8.4](http://ecma-international.org/ecma-262/5.1/#sec-8.4):

```
When a String contains actual textual data, each element is considered to be 
a single UTF-16 code unit.
```

So strings by default use UTF-16 encoding.

src/MergeContext.js
eduard-bagdasaryan: Done at eec9b9.

rousskov:
> When a String contains actual textual data, each element is considered to be a single UTF-16 code unit.

So the question becomes: Does `this._prMessage` contain "textual data" (from Javascript point of view)? It is clear that if _prMessage was initialized with (or set to) "textual data", then it will be interpreted as such here, and the length getter will return the number of characters rather than bytes. In that case, the validation method should work. However, if _prMessage was initialized with (or set to) raw bytes (without converting them or interpreting them as UTF), then .length will return the number of bytes and the String object contents will not correspond to the actual input if interpreted as UTF-16.

FYI: Your use of UTF (in the previously seemingly encoding-ignorant code) raised these concerns. We have been bitten by similar encoding problems in Daft (https://github.com/measurement-factory/daft/commit/da1f5f7). I just want to make sure we are not hitting or triggering them here again.

src/MergeContext.js
```diff
@@ -787,9 +787,24 @@ class PullRequest {
         return (this._rawPr.title + ' (#' + this._rawPr.number + ')' + '\n\n' + this._prBody()).trim();
     }
 
+    // Searches for the first non_ASCII_printable character and returns
+    // its position in the string. By default, JavaScript strings are
+    // represented as a sequence of UTF-16 characters.
+    _invalidCharacterPosition(str) {
+        const prohibitedCharacters = /[^\u0020-\u007f]+/; // allow non-special ASCII characters
```
rousskov: We need to find the first prohibited character. We do not care what follows. Let's remove `+` from the regex.

rousskov: I do not think we should allow the DEL character (ASCII code point 0x7F). Do you?

src/MergeContext.js
```diff
@@ -787,9 +787,24 @@ class PullRequest {
         return (this._rawPr.title + ' (#' + this._rawPr.number + ')' + '\n\n' + this._prBody()).trim();
     }
 
+    // Searches for the first non_ASCII_printable character and returns
+    // its position in the string. By default, JavaScript strings are
```
rousskov: Please remove the "By default..." sentence. The internal String "representation" (it is actually more like "interpretation" -- String storage is probably always an array of integers), especially the "default" one, seems irrelevant for this function interface.

src/MergeContext.js
rousskov: The code probably "works", but, AFAICT, we should be using `\u{}` characters and ES6 [regexes](https://mathiasbynens.be/notes/es6-unicode-regex) here. They are safer, especially when it comes to [character classes](https://mathiasbynens.be/notes/es6-unicode-regex#impact-character-classes).

eduard-bagdasaryan:
> So the question becomes: Does this._prMessage contain "textual data" (from Javascript point of view)?

The raw PR message (as well as other raw fields) are filled by 'JSON.parse(data)' in node-github library, where 'data' is what has come from the wire. I even have not found any attempts to make use of the received "Content-Type: application/json; charset=utf-8" header field in order to apply utf-8 encoding. Probably this is the default encoding for JSON parser. The [discussion](https://stackoverflow.com/questions/9254891/what-does-content-type-application-json-charset-utf-8-really-mean) has a reference to JSON [specs](https://www.ietf.org/rfc/rfc4627.txt), confirming that. I also performed a test, filling a file with a utf-8 two-byte symbol(c3a4) and running JSON.parse on it: the parsed string contained that symbol in UTF-16 ("00e4").
So, the library feeds us objects with strings, having converted multi-byte utf-8 sequences (if exist) into UTF-16 and the string length is the number of characters (not the number of received bytes).

src/MergeContext.js
eduard-bagdasaryan: I don't know why, but this code point escapes did not work in my tests (node v9). Moreover, I decided to use `\x` for matching, since it has the same effect and is shorter.

rousskov: > I don't know why, but this code point escapes did not work in my tests (node v9)

Just for the record, did you use the `/u` regex flag when testing `\u{}` code point expressions?

eduard-bagdasaryan: Well, I see: I looked [here](https://mathiasbynens.be/notes/javascript-escapes#unicode-code-point) when learning the syntax, where '/u' is not mentioned. Now it works.

rousskov: Right, that section talks about usage in strings, not regexes.

rousskov: > Now it works

... but the code is still missing `/u` flag for the regex. AFAICT, that may even be a source of a bug if our strings may contain Unicode:

> The u flag also affects negated character classes. For example, /[^a]/ is equivalent to /[\0-\x60\x62-\uFFFF]/, which would match any BMP symbol except a. But with the u flag, /[^a]/u matches the much bigger set of all Unicode symbols except a.

Please also use explicit brackets for Unicode points and, if possible, avoid leading zeros (e.g., `\u{20}`).

src/MergeContext.js
```diff
@@ -787,9 +787,24 @@ class PullRequest {
...
+        const match = prohibitedCharacters.exec(str);
+        return match ? match.index : -1;
+    }
+
     _prMessageValid() {
         const lines = this._prMessage().split('\n');
-        for (let line of lines) {
+        for (let i = 0; i < lines.length; ++i) {
+            const line = lines[i];
+            // this._prBody() already cared about removing trailing \r characters
```
```suggestion
            // this._prBody() already removed any trailing \r characters
```

Why are we talking about the _trailing_ `\r` here? A non-trailing `\r` in an `\r\n` sequence inside the text should also be ignored. Was it removed by the `.split()` call above?

eduard-bagdasaryan: this_prBody() may have a multiline string and it removes any \r, substituting \r+\n with \n. Since split() here uses \n as a separator, we can mistakenly assume that a legitimate preceding \r would force _prMessageValid() return false. The comment reminds us of that. Also, any \r (without following \n) within the text (body or title, no matter) would be treated as a forbidden symbol and mean an invalid message. Is it wrong?

src/MergeContext.js
eduard-bagdasaryan: Done at 4492ab51.

src/MergeContext.js
rousskov: Based on your description, the code works correctly, but the comment is lying -- `_prBody()` does not (and should not) remove trailing `\r` characters. I suggest adjusting the comment to better match reality:

```suggestion
            // _prBody() removed CRs in CRLF sequences; other CRs are banned
```

rousskov: I committed a better comment in 71def24.

PR #35, opened 2020-01-16
Ignore missing target_url in statuses

eduard-bagdasaryan:
Instead of marking the PR with M-failed-other, allow nil/empty
StatusCheck::targetUrl because GitHub does not require this field when
creating a new status for the staged commit, and Anubis does not use it
internally.

The target_url field is empty for the initial Jenkins 'pending' status,
when the test has been queued but not started yet.

### DIFF
--- a/src/MergeContext.js
+++ b/src/MergeContext.js
@@ -126,12 +126,14 @@ class StatusCheck
     constructor(raw) {
         assert(raw.context);
         assert(raw.state);
-        assert(raw.target_url);
         assert(raw.description);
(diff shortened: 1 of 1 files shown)


src/MergeContext.js
```diff
@@ -1040,7 +1051,7 @@ class PullRequest {
 
             const check = new StatusCheck({
                     state: "success",
-                    target_url: requiredPrStatus.targetUrl,
+                    target_url: requiredPrStatus.targetUrl(),
```
rousskov: Specifically, can we just _not_ copy the target URL if it was not provided?

PR #37, opened 2020-04-23
Added M-abandoned-staging-checks label

eduard-bagdasaryan:
If a PR changes after staging, its staged commit is classified as
'stale' and abandoned. Then either a new staged commit is created or the
control passes to another PR. In both cases, it may be unclear (from a
user's point of view) why the bot abandoned this seemingly good
(i.e., possibly green) staged commit. The new label allows to
distinguish such situations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,6 +163,9 @@ request state:
 * `M-merged`: The PR was successfully merged (and probably closed).
   The bot will not attempt to merge this PR again even if it is
   reopened. The bot never removes this label.
+* `M-abandoned-staging-checks`: The PR state changed, making the
+  still running CI tests stale/inapplicable. The bot removes this label
+  after creating a fresh stage commit for this PR.
 
 All labels, except `M-cleared-for-merge` and `M-merged`, are ignored by
 the bot itself! Humans may find them useful when determining the current
--- a/README.md
+++ b/README.md
@@ -145,9 +145,9 @@ request state:
   commits new code) or the staging_branch is unexpectedly modified. After
   labeling the PR, the bot ignores any failed or unfinished tests associated
   with that stale commit. This label was added so that the developer observing
-  the PR page on GitHub can be sure that Anubis had a good reason to ignore
-  failed or unfinished staging tests. The bot removes this (usually
-  short-lived) label after creating a fresh staging commit for the PR.
+  the PR page on GitHub knows why Anubis ignored (failed, unfinished, or
+  successful) staging tests. The bot removes this (usually short-lived) label
+  after creating a fresh staging commit for the PR.
 * `M-failed-staging-checks`: Essentially duplicates GitHub "red x" mark
   for the _staging commit_. The bot does not attempt to merge this PR
   again until a human decides that this problem is resolved and removes
--- a/README.md
+++ b/README.md
@@ -140,6 +140,14 @@ request state:
   mode (see `config::staged_run` and `config::guarded_run`). The bot
   removes this label when either the PR was successfully merged or its
   staging results are no longer fresh/applicable.
+* `M-abandoned-staging-checks`: Anubis discovered a stale _staging commit_.
+  This usually happens when either the PR state changes (e.g., the author
+  commits new code) or the staging_branch is unexpectedly modified. After
+  labeling the PR, the bot ignores any failed or unfinished tests associated
+  with that stale commit. This label was added so that the developer observing
+  the PR page on GitHub can be sure that Anubis had a good reason to ignore
+  failed or unfinished staging tests. The bot removes this (usually
+  short-lived) label after creating a fresh staging commit for the PR.
 * `M-failed-staging-checks`: Essentially duplicates GitHub "red x" mark
   for the _staging commit_. The bot does not attempt to merge this PR
   again until a human decides that this problem is resolved and removes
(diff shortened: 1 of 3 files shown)


src/MergeContext.js
```diff
@@ -269,6 +269,8 @@ class Label
 
     needsAdditionToGitHub() { return !this._presentOnGitHub && this._presentHere; }
 
+    onAddedToGitHub() { this._presentOnGitHub = true; }
```
rousskov, line 272: For consistency with other similar methods:

```suggestion
    markAsAdded() { this._presentOnGitHub = true; }
```

If it is not difficult, let's also call this new method after pushing any label, probably inside _addToGitHub(). This will require passing the label to that method (as opposed to just the label name), but that is OK.

src/MergeContext.js
```diff
@@ -287,11 +289,23 @@ class Labels
...
-            this._labels.push(new Label(name, false));
+        else {
+            label = new Label(name, false);
+            this._labels.push(label);
+        }
+        return label;
+    }
+
+    // adds a label and immediately applies it to GitHub
+    async addAndPush(name) {
```
rousskov, line 303: Technically, the regular/old add() also pushes, just not immediately -- the push is delayed. The fact that add() delays pushing is mostly irrelevant to its callers -- everything would probably "work" even if add() was immediately pushing. That is why it is OK to leave add() name as it is IMO. Let's clarify the difference of this new method though:

```suggestion
    async addImmediately(name) {
```

README.md
```diff
@@ -163,6 +163,10 @@ request state:
 * `M-merged`: The PR was successfully merged (and probably closed).
   The bot will not attempt to merge this PR again even if it is
   reopened. The bot never removes this label.
+* `M-abandoned-staging-checks`: The PR state changed, making the
```
rousskov, line 166: Just a question: Would this label be added if the PR state remains the same but the state of the `auto` branch changes? For example, if another PR (or a human) force-pushes their own commit to the auto branch? I would like to polish this wording, but I want to make sure I understand what triggers the label.

src/MergeContext.js

src/MergeContext.js

README.md
```diff
@@ -140,6 +140,14 @@ request state:
   mode (see `config::staged_run` and `config::guarded_run`). The bot
   removes this label when either the PR was successfully merged or its
   staging results are no longer fresh/applicable.
+* `M-abandoned-staging-checks`: Anubis discovered a stale _staging commit_.
+  This usually happens when either the PR state changes (e.g., the author
+  commits new code) or the staging_branch is unexpectedly modified. After
+  labeling the PR, the bot ignores any failed or unfinished tests associated
+  with that stale commit. This label was added so that the developer observing
+  the PR page on GitHub can be sure that Anubis had a good reason to ignore
+  failed or unfinished staging tests. The bot removes this (usually
```
eduard-bagdasaryan, line 149:
```suggestion
  previous (failed, unfinished or succeeded) staging tests. The bot removes this (usually
```

README.md
```diff
@@ -140,6 +140,14 @@ request state:
   mode (see `config::staged_run` and `config::guarded_run`). The bot
   removes this label when either the PR was successfully merged or its
   staging results are no longer fresh/applicable.
+* `M-abandoned-staging-checks`: Anubis discovered a stale _staging commit_.
+  This usually happens when either the PR state changes (e.g., the author
+  commits new code) or the staging_branch is unexpectedly modified. After
```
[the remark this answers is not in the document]
rousskov: > there is nothing extraordinary in modifying the PR branch during staging

Agreed, but that is not the event the "unexpected" applies to. The text talks about unexpected modification of the staging branch, not the PR branch. When we talk about the staging branch, we expect that only Anubis can modify it and that once Anubis stages a PR it does not stage any other PR (without having to go through the labeling code of the staged PR), right?

Are there any _expected_ modifications of the _staging_ branch that would still result in a `M-abandoned-staging-checks` label?

README.md
rousskov: Good point! Addressed in aa1e507?

README.md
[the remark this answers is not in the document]
eduard-bagdasaryan: > The text talks about unexpected modification

Ah, I see. Yes, such external modifications of auto branch (performed not by the bot) will result in `M-abandoned-staging-checks`. We should leave the wording it as is then.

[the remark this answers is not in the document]
rousskov: Great, the ball is on your side then.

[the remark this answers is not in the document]
eduard-bagdasaryan: > Are there any expected modifications of the staging branch 

If by 'expected' we mean 'performed by Anubis' - I think such modifications exist. For example, an exception/assertion in the bot leads to the PR abandoning and switching to another PR, staging it, and modifying auto branch. However, we can treat this emergency situation as 'unexpected' (though Anubis is aware of it).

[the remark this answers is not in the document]
rousskov: > However, we can treat this emergency situation as 'unexpected' (though Anubis is aware of it).

Yes, failed assertions can be classified as unexpected events, regardless of how they are handled. However, I am not sure they are actually relevant here because, after #36, when there is an assertion, there will be an `M-failed-staging-other` or `M-failed-other` label _instead_ of the `M-abandoned-staging-checks` label, right? In other words, the new `M-abandoned-staging-checks` label cannot appear for PR X _because_ Anubis asserted while processing PR X, correct?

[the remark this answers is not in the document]
eduard-bagdasaryan: > after #36, when there is an assertion, there will be an M-failed-staging-other or M-failed-other label instead of the M-abandoned-staging-checks label,

Agreed, this is not relevant - `M-abandoned-staging-checks` will not appear in a case of assertion because the only place it is applied now - after the discovery of a stale PR. For now (i.e, without PR36 changes), assertions will result in `M-failed-other`.

PR #40, opened 2020-10-28
Force waiting for automated merge

eduard-bagdasaryan:
Each GitHub PR has a manual merge button on the GitHub PR page,
which becomes green as soon as all preconditions (mandatory PR checks)
are satisfied. At that time, some admins may click on that button,
resulting in wrong commit messages and missing auto tests (at best).

To reduce the probability of such accidents, Anubis now automatically
adds a "waiting for automated merge" check, marked as required on
GitHub.

* For the staged commit, the check is satisfied immediately before
  the attempted merge (so that GitHub is happy about all the required
  status checks).

* For the PR, the check is satisfied only after Anubis merges the
  staged commit.

### DIFF
--- a/src/MergeContext.js
+++ b/src/MergeContext.js
@@ -160,16 +160,16 @@ class StatusChecks
         assert(expectedStatusCount !== undefined);
         assert(expectedStatusCount !== null);
         assert(context);
-        this.expectedStatusCount = expectedStatusCount;
+        this._expectedStatusCount = expectedStatusCount;
         this.context = context;
-        this.requiredStatuses = [];
+        this._requiredStatuses = [];
         this.optionalStatuses = [];
     }
 
     addRequiredStatus(requiredStatus) {
         assert(requiredStatus);
         assert(!this.hasStatus(requiredStatus.context));
-        this.requiredStatuses.push(requiredStatus);
+        this._requiredStatuses.push(requiredStatus);
     }
 
     addOptionalStatus(optionalStatus) {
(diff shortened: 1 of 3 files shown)


src/MergeContext.js
```diff
@@ -201,10 +201,40 @@ class StatusChecks
         this.addRequiredStatus(new StatusCheck(raw));
     }
 
+    hasAutomatedMergeStatus(state, description) {
+        return this._requiredStatuses.some(el =>
+                el.context.trim() === Config.automatedMergeStatusContext() &&
+                el.state === state &&
+                el.description === description);
```
rousskov, line 208: If possible, reorder to place faster and more selective conditions first:

```suggestion
                el.description === description &&
                el.state === state &&
                el.context.trim() === Config.automatedMergeStatusContext();
```

src/MergeContext.js
```diff
@@ -1166,6 +1237,8 @@ class PullRequest {
 
         assert(!this._stagingBanned);
 
+        await this._setAutomatedMergeStatusForStaged("success");
```
rousskov, line 1240: Since we are setting this a bit prematurely/optimistically, it would be nice to make this one `success (will merge SHA)` and then, if the status may become visible anywhere after we successfully merge (as discussed in another change request), change that to `success (M-merged as SHA)`.

rousskov (changes requested):

eduard-bagdasaryan (commented):
Done (28113c5, 3cfb8ba).

Issue #41, opened 2021-01-25
Limit PR processing time


eduard-bagdasaryan:
We need a mechanism to prevent Anubis stalling in cases when GitHub communication is too slow or an API method got stuck. Probably we should introduce a time limit allocated for each PR and throw an 'unexpected error' when a PR exceeds this limit.

We experienced this problem when Squid PR756 got stuck. According to the logs, the bot completed compareCommits() but did not proceed any further with this PR (no more lines related to this PR were logged). On the other hand, the bot process was still executing, updating the log with GitHub status events. I speculate that it could be getRawPR() method which caused stalling, but unfortunately we lack any debugging lines at the beginning of an API method. I think we would need to add them too.

PR #42, opened 2021-02-01
Abort stalled GitHub transactions

eduard-bagdasaryan:
The bot may wait for a GitHub response for too long (several hours),
reacting to notifications but stalling progress on all PRs. We need a
timeout to avoid such delays. Fortunately, the octokit/rest library
provides the 'timeout' parameter (which defaults to no timeout). A
timed-out API call returns a 504 error, and we handle that correctly.

TODO: Eventually, we may add a more general "PR processing" timeout, but
this immediate solution is too simple to pass up.

Addresses issue #41.

### DIFF
--- a/src/Config.js
+++ b/src/Config.js
@@ -113,8 +113,9 @@ class ConfigOptions {
 
     copiedDescriptionSuffix() { return " (copied from PR by Anubis)"; }
 
-    // GitHub request timeout, ms
-    requestTimeout() { return 60000; }
+    // GitHub transaction timeout, ms
+    // GitHub library default is 0 which is interpreted as infinity (i.e. no timeout)
+    requestTimeout() { return 3 * 60000; }
 }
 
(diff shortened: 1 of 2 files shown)


src/Config.js
```diff
@@ -112,6 +112,9 @@ class ConfigOptions {
     approvalContext() { return "PR approval"; }
 
     copiedDescriptionSuffix() { return " (copied from PR by Anubis)"; }
+
+    // GitHub request timeout, ms
+    requestTimeout() { return 60000; }
```
rousskov, line 117: If this applies to _all_ bot communications with GitHub:

```suggestion
    // GitHub transaction timeout, ms
    // GitHub library default is zero which is interpreted as infinity (i.e. no timeout)
    githubTimeout() { return 60000; }
```

rousskov (approved):
Approved with changes. I also adjusted the PR title/description. Please check before merging.

PR #43, opened 2021-02-04
Removed the need for humans to manage M-failed-staging-checks

eduard-bagdasaryan:
The bot relied on a human to delete some stale M-failed-staging-checks
labels but also ignored stale M-failed-staging-checks when the restarted
checks succeeded. The latter led to merged PRs still carrying a
M-failed-staging-checks label. While fixing that specific problem, we
realized that the bot can manage this label exclusively, without any
human help (and adjusted the code and documentation accordingly).

Also polished _removeTemporaryLabels() documentation even though the
now-explicitly-covered "labels that are also removed by humans" case
is currently not represented by any label.

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,8 +150,11 @@ request state:
   after creating a fresh staging commit for the PR.
 * `M-failed-staging-checks`: Essentially duplicates GitHub "red x" mark
   for the _staging commit_. The bot does not attempt to merge this PR
-  again until a human decides that this problem is resolved and removes
-  the label manually.
+  again until either of two conditions:
+  * A human decides that this problem is resolved and removes the label
+  manually.
+  * This staging commit checks restart and succeed. In this case
+  Anubis proceeds with the PR, removing the label.
 * `M-failed-staging-other`: A fatal PR-specific error that occurred while
   waiting for staging checks but was not classified as
   `M-failed-staging-checks`. It is probably necessary to consult CI logs to
--- a/README.md
+++ b/README.md
@@ -150,11 +150,11 @@ request state:
   after creating a fresh staging commit for the PR.
 * `M-failed-staging-checks`: Essentially duplicates GitHub "red x" mark
   for the _staging commit_. The bot does not attempt to merge this PR
-  again until either of two conditions:
+  again until at least one of the following events happens:
   * A human decides that this problem is resolved and removes the label
-  manually.
-  * This staging commit checks restart and succeed. In this case
-  Anubis proceeds with the PR, removing the label.
+    manually.
+  * The failed checks for the staging commit are restarted or removed.
+  * The staging commit becomes stale. See `M-abandoned-staging-checks`.
 * `M-failed-staging-other`: A fatal PR-specific error that occurred while
   waiting for staging checks but was not classified as
   `M-failed-staging-checks`. It is probably necessary to consult CI logs to
--- a/README.md
+++ b/README.md
@@ -148,13 +148,10 @@ request state:
   the PR page on GitHub knows why Anubis ignored (failed, unfinished, or
   successful) staging tests. The bot removes this (usually short-lived) label
   after creating a fresh staging commit for the PR.
-* `M-failed-staging-checks`: Essentially duplicates GitHub "red x" mark
-  for the _staging commit_. The bot does not attempt to merge this PR
-  again until at least one of the following events happens:
-  * A human decides that this problem is resolved and removes the label
-    manually.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -150,8 +150,11 @@ request state:
   after creating a fresh staging commit for the PR.
 * `M-failed-staging-checks`: Essentially duplicates GitHub "red x" mark
   for the _staging commit_. The bot does not attempt to merge this PR
-  again until a human decides that this problem is resolved and removes
-  the label manually.
+  again until at least one of the following events happens:
+  * A human decides that this problem is resolved and removes the label
+    manually.
```
rousskov, line 155: How is this manual removal relevant? Would not Anubis recheck the still-failed status checks for the staged commit and re-add the label?

rousskov (changes requested):
The initial changes looked very strange to me. I tried to adjust the code, but I am still not sure the current version is what we want. Please check my changes and answer one specific question in the inlined change request.

README.md
eduard-bagdasaryan: >  Would not Anubis recheck the still-failed status checks

It will recheck and re-add this label if nothing changes in this PR: there is no use to drop this label in this case. The user should fix the PR manually (i.e., commit some fixes) and only then drop the label.

rousskov: > there is no use for a human to drop this label in this case

Agreed.

> The user should fix the PR manually (i.e., commit some fixes) and only then drop the label.

Hm... That does not sound like a valid case to me because if the user commits new fixes, Anubis will know that the staged commit is stale and will drop the label automatically, right?

Similarly, if the user restarts tests and they succeed (without any new branch commit), then Anubis will also notice the lack of failures and will remove the label automatically, right?

I am trying to find a case where a human must delete the label manually to ensure progress. I cannot find it in the current branch code. If such a case does not exist now, then either the branch code is buggy or we need to fix the documentation to remove the human-focused case.

eduard-bagdasaryan: > if the user commits new fixes, Anubis will know that the staged commit is stale and will drop the label automatically

No, it will not drop the label automatically. I guess the intention was to allow the user to do whatever required with this PR postponing Anubis interaction until the user decides that the PR is fixed and drops the label.

>  if the user restarts tests and they succeed (without any new branch commit), then Anubis will also notice the lack of failures and will remove the label automatically

This is what this PR is attempting to fix.

rousskov: > Alex: if the user commits new fixes, Anubis will know that the staged commit is stale and will drop the label automatically

> Eduard: No, it will not drop the label automatically

Why not? In other words, what PR branch code will add the label (after calling _removeTemporaryLabels()) or prevent _removeTemporaryLabels() from being called in the first place? Note that the "Do not vainly recreate staged..." code that adds this label is only executed when the staged commit is fresh. We are discussing a case where the staged commit becomes stale because of the new committed fixes.

rousskov: > This is what this PR is attempting to fix.

Yes, and I assume that it does fix that case. So that case is removed from the candidates in my search for a case where a human should remove the label.

eduard-bagdasaryan: > Why not?

In my statement, I meant master code.

> what PR branch code will add the label

The label is added in two cases:
* in _processStagingStatuses() before attempting to merge the PR. We do not check the staging commit freshness here as in _loadPrState() (the bot states that a commit is fresh if either it has been just created or already checked in _loadPrState() - at the beginning of the PR processing cycle).
* in _loadPrState(), when the bot discovers that the PR is fresh and has broken staging checks. 

With PR changes (including the latest changes), this label will be removed in _removeTemporaryLabels(), running before  _loadPrState(). So if the PR is not fresh, it will not be set in _loadPrState() and it will not be set in _processStagingStatuses() because we force 'brewing'. In other words, the bot will not add the label in this case.

rousskov: The implications of your last comment seem to match my understanding of the PR state. Focusing on this PR code, do you agree that we should remove the "human" case/bullet from the documentation update that piece this change request is attached to? I will adjust the documentation around that removal.

eduard-bagdasaryan: We should remove this bullet to match the PR code.

rousskov: Done in ca328b8.

rousskov (approved):
I think the PR now addresses the problem correctly. Please check that I did not miss anything. If you are happy with the code, please merge and proceed with deploying.

rousskov:
I also adjusted PR title/description to match the current changes. Please check that as well.

PR #44, opened 2021-03-12
Removed GitHub master commits search optimization

eduard-bagdasaryan:
Github Search API has a custom rate limit, which is 30 requests per
minute for our token authorization. This limitation did not matter when
there were few open PRs. Now, as the PR list has grown, the limit was
reached on every pass (according to the bot logs) and several
unfortunate (random) PRs were constantly marked with M-failed-other.

Fortunately, we can do without this advanced Search API and use the
standard GitHub 'List commits' API. There is a couple of disadvantages
with this approach:

1. It does not allow filtering by searching for the PR number
in commit messages. As a result, if previously the received list of
commits was always empty (except rare cases when somebody has
merged PR by hand) but now it will usually contain several PRs and
the bot itself will filter the received results.

2. It is 15-20% slower, which may be a consequence of (1) due to the
increased data traffic. However, taking into account the 30/minute
limitation, the final PR scan time would be almost the same (with the
difference that Anubis would wait until GitHub resets its counters).

So, since the benefits of the advanced search are small versus
implementation complexity, we decided to drop it for good.

### DIFF
--- a/src/GitHubUtil.js
+++ b/src/GitHubUtil.js
@@ -441,24 +441,6 @@ function getUserEmails() {
     });
 }
 
-function searchCommits(query) {
-    let params = {};
-    params.q = query;
-    params.sort = 'committer-date';
-    return new Promise( (resolve, reject) => {
-      GitHub.authenticate(GitHubAuthentication);
-      GitHub.search.commits(params, (err, res) => {
-          if (err) {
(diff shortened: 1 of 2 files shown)


rousskov (approved):
LGTM, thank you.

We also considered dropping the entire _mergedSomeTimeAgo(). We may do that later if the remaining search proves to be problematic.

PR #45, opened 2021-03-12
Speed up GitHub search for a merged PR

eduard-bagdasaryan:
The "author" filter was forgotten since we started searching in f7acf93.


rousskov (approved):
LGTM. Glad you noticed this bug!

PR #46, opened 2021-03-17
Do not drop M-failed-staging-checks prematurely

eduard-bagdasaryan:
Before 1279629, the label prevented the unchanged PR progress, avoiding
useless staging commit recreation. Since 1279629 this is broken because
this label is dropped (along with other labels) in
_removeTemporaryLabels() and never set again.

We don't need to remove this label when:

* The staged commit is fresh and its statuses are still failed.

* The PR does not have a staged commit now (but had it with failed
statuses some time ago, as M-failed-staging-checks indicates).
The abandoned staged commit is still fresh and we do not
need to vainly recreate it in this case (and possibly
bringing about a livelocking situation with two such PRs).

We still need to remove this label when:

* The staged commit is stale and is being abandoned.

* The staged commit is fresh but its staging statuses succeeded
(after previous failure(s)).

* The PR does not have a staged commit now (but had it with failed
statuses some time ago, as M-failed-staging-checks indicates).
The abandoned staged commit became stale since then (e.g., some new
branch commits were pushed or the statuses of the abandoned commit
changed). We should proceed with this PR (removing the label).

In order to distinguish these cases, we need to get PR's abandoned stage
commits and calculate their freshness.  A new getIssueEvents() API
allows to get all PR-associated (i.e., 'referenced') commits from the PR
event history, while the adjusted compareCommits(..., diffFormat) allows
to determine whether the abandoned staged commit is fresh.


src/MergeContext.js
```diff
@@ -967,23 +982,27 @@ class PullRequest {
 
         if (this._stagedPosition.merged()) {
             this._log("already merged into base some time ago");
+            // just in case - it should be false here
+            this._stagingFailedSomeTimeAgo = false;
```
rousskov, line 986:
```suggestion
            // cleanup if this interrupted merge left a stale label behind
            this._stagingFailedSomeTimeAgo = false; // probably already false
```

src/MergeContext.js
```diff
@@ -1295,6 +1314,9 @@ class PullRequest {
             else
                 this._removePositiveStagingLabels();
 
+            // some of these labels may be set already in the exception
+            this._addPrStateLabels();
```
rousskov, line 1318: AFAICT, to do this addition here, you have to move `if (!this._labels)` from below to here first. The "report this..." comment attached to that `if` statement should stay where it is now, of course.

rousskov, line 1318: I think we should delete/inline the used-once _addPrStateLabels()  method here. We are kind of working in a dangerous spot here -- handling any exception -- and I am worried that we will forget about it and start adding advanced code into that method if it will be called from somewhere else as well. Inlining will also clarify our logic, at least for now.

src/MergeContext.js
```diff
@@ -954,8 +959,18 @@ class PullRequest {
         return false;
     }
 
+    /// whether this PR has or had a staged commit with failed checks
+    _stagedStatusesFailed() {
+        return this._stagedStatuses === null ?
+            this._stagingFailedSomeTimeAgo : this._stagedStatuses.failed();
```
rousskov, line 965: This looks a bit inconsistent with the other code -- sometimes we update _stagingFailedSomeTimeAgo after _getStagingStatuses() and sometimes we do not. There are also calls to this._stagedStatuses.failed() that do not go through this function.

I think we should rename _stagingFailedSomeTimeAgo to _stagingFailed and make sure it is _always_ in sync with _stagedStatuses (when they are available). After that, we can drop this function and simply check _stagingFailed directly.

rousskov (changes requested):
I think you are on the right track here, but we need to tighten it up a little. If any of my suggested source code comments do not match reality, please let me know.

src/MergeContext.js
eduard-bagdasaryan: Renamed.

> always in sync with _stagedStatuses

To keep such synchronization I moved this flag assignment into _getStagingStatuses().

eduard-bagdasaryan (commented):
Done at b38e37f, 94aeaf0.

src/MergeContext.js
```diff
@@ -954,36 +964,77 @@ class PullRequest {
...
+            // cleanup if this interrupted merge left a stale label behind
+            this._stagingFailed = false; // probably already false
             this._enterMerged();
             return;
         }
 
         if (!(await this._stagedCommitIsFresh())) {
             await this._labels.addImmediately(Config.abandonedStagingChecksLabel());
+            // forget label-based info derived from a now-stale staged commit
+            this._stagingFailed = false;
```
rousskov, line 1028: The source of that stale info (a label or commit analysis) seems irrelevant here. Only the fact that we do not have a fresh staged commit anymore is relevant.

```suggestion
            // we have no fresh staged commit
            this._stagingFailed = false; // may already be false
```

src/MergeContext.js
```diff
@@ -954,36 +964,77 @@ class PullRequest {
...
+            await this._enterBrewing();
             return;
         }
 
         assert(this._stagedPosition);
 
         if (this._stagedPosition.merged()) {
             this._log("already merged into base some time ago");
+            // cleanup if this interrupted merge left a stale label behind
+            this._stagingFailed = false; // probably already false
```
rousskov, line 1020: Just to avoid the implication that we are describing the `if` statement condition here:
```suggestion
            // cleanup in case this interrupted merge left a stale label behind
            this._stagingFailed = false; // probably already false
```

However, what if that label was not actually stale? We do accept that possibility in the _mergedSomeTimeAgo() case above... I suggest that we simplify and do nothing special for any merged() PRs. If they had the label and where somehow (manually?) merged despite it, keep the label. That would mean removing this comment/code _and_ the new comment in the _mergedSomeTimeAgo() case.

src/MergeContext.js
```diff
@@ -1182,10 +1233,10 @@ class PullRequest {
         assert(Config.githubUserName());
     }
 
+
```
rousskov, line 1236: Undo an out-of-scope change?

src/MergeContext.js
```diff
@@ -695,6 +702,7 @@ class PullRequest {
         let labels = await GH.getLabels(this._prNumber());
         assert(!this._labels);
         this._labels = new Labels(labels, this._prNumber());
+        this._stagingFailed = this._labels.has(Config.failedStagingChecksLabel());
```
rousskov, line 705: Why do we need to set this._stagingFailed based on the label at all? Why not always compute its value from scratch, based on staged commit history/state? AFAICT, this complication is rooted in a case where there is no staged commit anymore, but our _recalculateStagingFailed() handles that exact situation already! Perhaps this is just an optimization that is meant to reduce the number of _recalculateStagingFailed() calls?

rousskov (changes requested):
We are making progress, but these complex changes and concerns will require at least one more iteration. I hope the updated field definition (despite being complex) will help simplify these changes.

eduard-bagdasaryan (commented):
Done (e380e54).

src/MergeContext.js
eduard-bagdasaryan: After some consideration, I decided to not compute _stagingFailed from the label as you suggested. Done at cde4f45.

src/MergeContext.js
```diff
@@ -954,12 +968,47 @@ class PullRequest {
         return false;
     }
 
+    async _getMergeCommit() {
+        const mergeSha = await GH.getReference("pull/" + this._prNumber() + "/merge");
+        return await GH.getCommit(mergeSha);
+    }
+
+    // recalculates this._stagingFailed value for PRs without staging commit
```
rousskov, line 976:
```suggestion
    // recalculates this._stagingFailed value for PRs without a staged commit
```

src/MergeContext.js
```diff
@@ -954,12 +968,47 @@ class PullRequest {
...
+    async _recalculateStagingFailed() {
+        assert(!this._stagedSha());
+
+        const allEvents = await GH.getIssueEvents(this._prNumber());
+        // we consider all commits created by the bot user and referencing this PR as staged commits
```
rousskov, line 981:
```suggestion
        // staging events are PR events where the bot user created a commit referencing this PR
```

src/MergeContext.js
```diff
@@ -954,12 +968,47 @@ class PullRequest {
...
+        let stagedEvents = allEvents.filter(ev => ev.event === "referenced" && ev.actor.login === Config.githubUserLogin());
```
rousskov, line 982: I assume there are many PR events that are not related to commit creation. Let's assume that future Anubis versions will create such events as well, even if they do not do that today. Should not our filter expression discard events that are not related to a commit creation? Or does `ev.event === "referenced"` already do that reliably? In other words, does `ev.event === "referenced"` imply that `ev` is a commit creation event and not some other event that "referenced" something?

```suggestion
        let stagingEvents = allEvents.filter(ev => ev.event === "referenced" && ev.actor.login === Config.githubUserLogin());
```

src/MergeContext.js
```diff
@@ -954,12 +968,47 @@ class PullRequest {
...
+        if (!stagedEvents.length)
+            return;
+
+        // just in case: events should be already sorted by date
```
rousskov, line 986:
```suggestion
        // just in case: events should be in chronological order already
```

src/MergeContext.js
```diff
@@ -954,12 +968,47 @@ class PullRequest {
...
+        stagedEvents = stagedEvents.sort((ev1, ev2) => Date.parse(ev1.created_at) - Date.parse(ev2.created_at));
+        const lastStaged = stagedEvents[stagedEvents.length - 1];
+        const mergeCommit = await this._getMergeCommit();
+        const mergeCommitCreatedAt = Date.parse(mergeCommit.author.date);
+        const stagedCommitCreatedAt = Date.parse(lastStaged.created_at);
```
rousskov, line 991: I think it is helpful to keep the "last" and "event" parts because, per our definitions, there can be only one staged commit at any given time (but many staged commit creation events which we call "staging events").

```suggestion
        const lastStagingEventCreatedAt = Date.parse(lastStagingEvent.created_at);
```

src/MergeContext.js
```diff
@@ -954,12 +968,47 @@ class PullRequest {
...
+        assert(mergeCommitCreatedAt);
+        assert(stagedCommitCreatedAt);
+        // whether the merge commit is fresher than the last staged commit
+        if (stagedCommitCreatedAt < mergeCommitCreatedAt)
+            return false;
```
rousskov, line 996: The "false" return value seems to be unnecessary/undefined/unused. Should not we set the filed we were asked to recalculate instead of returning false? Let's also clarify why we do not care about the equality case. Adjust as needed, of course.

```suggestion
        // the staging event for merge commit X is always created after X so we do not
        // need to deal with the lastStagingEventCreatedAt == mergeCommitCreatedAt case
        assert(lastStagingEventCreatedAt != mergeCommitCreatedAt);
        if (lastStagingEventCreatedAt < mergeCommitCreatedAt) {
            // merge commit created after the last staging event
            this._stagingFailed = true;
            return; 
        }
```

src/MergeContext.js
```diff
@@ -954,12 +968,47 @@ class PullRequest {
...
+
+        // recalculates this._stagingFailed value
```
rousskov, line 998:
```suggestion
        // Merge commit created before the last staging event. Thus, we can
        // (re)calculate this._stagingFailed value based on that event commit SHA
```

src/MergeContext.js
```diff
@@ -954,12 +968,47 @@ class PullRequest {
...
+
+    // recalculates this._stagingFailed value for PRs without staging commit
+    async _recalculateStagingFailed() {
+        assert(!this._stagedSha());
+
+        const allEvents = await GH.getIssueEvents(this._prNumber());
+        // we consider all commits created by the bot user and referencing this PR as staged commits
+        let stagedEvents = allEvents.filter(ev => ev.event === "referenced" && ev.actor.login === Config.githubUserLogin());
+        if (!stagedEvents.length)
+            return;
```
rousskov, line 984: Should not we recalculate in this case, as promised by the function name/description?

```suggestion
        if (!stagingEvents.length) {
            this._stagingFailed = false;
            return;
        }
```

src/MergeContext.js
```diff
@@ -954,12 +968,47 @@ class PullRequest {
...
 
+    async _getMergeCommit() {
+        const mergeSha = await GH.getReference("pull/" + this._prNumber() + "/merge");
+        return await GH.getCommit(mergeSha);
+    }
+
+    // recalculates this._stagingFailed value for PRs without staging commit
+    async _recalculateStagingFailed() {
+        assert(!this._stagedSha());
+
```
rousskov, line 979: Perhaps we should confirm the intent to recalculate by erasing the old value?

```suggestion

    this._stagingFailed = null;

```

Should not we also init with `null` in the class constructor because, at that time, we do not know the correct value?

rousskov (changes requested):
This is a partial review. It probably makes sense to resume this work only after the other active Anubis PR is merged so that you do not have to merge this code many times.

src/MergeContext.js
eduard-bagdasaryan: >  Or does ev.event === "referenced" already do that reliably?

According to the [documentation](https://docs.github.com/en/developers/webhooks-and-events/events/issue-event-types#referenced), the "referenced' attribute means "The issue was referenced from a commit message.". This should be quite reliable - as long as only Anubis creates commits referencing PRs.

eduard-bagdasaryan (commented):
Done (up to 051820a).

src/MergeContext.js
```diff
@@ -954,12 +962,50 @@ class PullRequest {
...
+        stagingEvents = stagingEvents.sort((ev1, ev2) => Date.parse(ev1.created_at) - Date.parse(ev2.created_at));
+        const lastStagingEvent = stagingEvents[stagingEvents.length - 1];
+        const mergeCommit = await this._getMergeCommit();
+        const mergeCommitCreatedAt = Date.parse(mergeCommit.author.date); // for merge commits author.date and committer.date should be the same
+        const lastStagingEventCreatedAt = Date.parse(lastStagingEvent.created_at);
+        assert(mergeCommitCreatedAt);
+        assert(lastStagingEventCreatedAt);
+        // the staging event for merge commit X is always created after X so we do not
+        // need to deal with the lastStagingEventCreatedAt == mergeCommitCreatedAt case
+        assert(lastStagingEventCreatedAt != mergeCommitCreatedAt);
```
rousskov, line 990: GitHub event times have a second resolution. Could not the second-resolution timestamps be the same even though the two events happened at "slightly" different times, leading to an assertion here?

P.S. Do we need to use `!==` here?

src/MergeContext.js
```diff
@@ -954,12 +962,50 @@ class PullRequest {
...
+        if (lastStagingEventCreatedAt < mergeCommitCreatedAt)
+            return; // merge commit created after the last staging event
+
+        // Merge commit created before the last staging event.
+        // Thus, we can get statuses of that event commit SHA.
```
rousskov, line 995: Should we say "we can use statuses..." instead of "we can get statuses..."? We can get statuses of any tested commit, right? We just do not want use the old/irrelevant statuses, right?

src/MergeContext.js
```diff
@@ -954,12 +962,50 @@ class PullRequest {
...
+        const mergeCommit = await this._getMergeCommit();
+        const mergeCommitCreatedAt = Date.parse(mergeCommit.author.date); // for merge commits author.date and committer.date should be the same
+        const lastStagingEventCreatedAt = Date.parse(lastStagingEvent.created_at);
+        assert(mergeCommitCreatedAt);
+        assert(lastStagingEventCreatedAt);
+        // the staging event for merge commit X is always created after X so we do not
+        // need to deal with the lastStagingEventCreatedAt == mergeCommitCreatedAt case
+        assert(lastStagingEventCreatedAt != mergeCommitCreatedAt);
+        if (lastStagingEventCreatedAt < mergeCommitCreatedAt)
+            return; // merge commit created after the last staging event
```
rousskov, line 992: Why is that a problem? In other words, we cannot we just use the lastStagingEvent even if there was a more recent merged commit? After all, we are not looking for a fresh/current staged commit here but an "abandoned" one... Why do we (implicitly) define "abandoned" this way?

src/GitHubUtil.js
```diff
@@ -86,6 +86,24 @@ function getLabels(prNum) {
...
+            if (err) {
+                reject(new ErrorContext(err, getIssueEvents.name, params));
+                return;
+            }
+            res = await pager(res);
+            const result = res.data.length;
+            logApiResult(getIssueEvents.name, params, result);
+            resolve(res.data);
+        });
+    });
```
rousskov, line 104: The above will need to be adjusted to use rateLimitedPromise(), right?

rousskov (commented):
Just another partial review, with mostly questions rather than "true" change requests.

src/MergeContext.js
eduard-bagdasaryan: We look for an abandoned staged commit (i.e., its statuses) only if there is no newer merge commit - that is how our abandonedStagedStatuses defined. A newer merge commit means that we do not need the abandonedStagedStatuses check at all in _checkStagingPreconditions() - there should be no livelocking scenario (which we try to avoid) in this case.

However, the current approach with dates comparison is flawed, IMO, because staged commit uses merge commit data, but the merge commit itself can be already stale by that time - GitHub is free to create a new one anytime the base or issue branch changes. In other words, theoretically there can be a situation when lastStagingEventCreatedAt > mergeCommitCreatedAt, but merge commit is fresher that the staged commit. I think we need another approach here - the one that we already use in _stagedCommitIsFresh(), comparing tree SHAs.

src/MergeContext.js
eduard-bagdasaryan: You are right that we can get statuses of any commit - as far as we have its SHA. Probably we can drop this comment at all - especially if we start comparing tree SHAs, as I suggested above.

src/MergeContext.js
eduard-bagdasaryan: > the second-resolution timestamps be the same

That is possible, I think, because merge commits and staged commits are created independently by GitHub and Anubis.

eduard-bagdasaryan (commented):
Answered questions. I'll post the required changes soon.

src/MergeContext.js
```diff
@@ -954,12 +962,50 @@ class PullRequest {
...
+        // staging events are PR events where the bot user created a commit referencing this PR
+        let stagingEvents = allEvents.filter(ev => ev.event === "referenced" && ev.actor.login === Config.githubUserLogin());
+        if (!stagingEvents.length)
+            return;
+
+        // just in case: events should be already in chronological order already
+        stagingEvents = stagingEvents.sort((ev1, ev2) => Date.parse(ev1.created_at) - Date.parse(ev2.created_at));
+        const lastStagingEvent = stagingEvents[stagingEvents.length - 1];
+        const mergeCommit = await this._getMergeCommit();
+        const mergeCommitCreatedAt = Date.parse(mergeCommit.author.date); // for merge commits author.date and committer.date should be the same
```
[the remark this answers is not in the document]
rousskov: > is there any reason why they should different

Just my personal (and possibly wrong) interpretation of what "merge commit" is. To me, that commit brings PR changes (authored _some time ago_) into the target repository branch. Thus, I would expect the author date to reflect something from the PR past rather than current time. However, I would not be surprised if somebody who knows a lot more about the mechanics of these merge commits would explain why the dates must be the same. It does not seem natural/intuitive to me, but sometimes the best approach is not natural or intuitive, especially from an outsider point of view.

To avoid misunderstanding, the problem here is _not_ what GitHub (currently) does.

* One problem is how we describe what we think it (currently) does. We describe it as if it is some self-evident truth or obvious principle, but we seem to be unable to substantiate that.
* Another problem is that given a choice between two dates, if we need to know when the merge commit was _created_ at, the committer date seems to be a more natural choice (when we look at commits where the dates differ, it is the committer date that we would pick). This is especially true if we would prefer to use the updated/new date if the merge commit was modified for some reason (but still represents a true merge commit). I know we do not expect such modifications, but let's assume, for the sake of the argument, that they happened to a merge commit -- which date would we pick then?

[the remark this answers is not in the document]
eduard-bagdasaryan: > the mechanics of these merge commits

I don't think that there is some special magic (or hidden mechanics) in this merge commits - except the fact that they are automatically generated by GitHub. For example, I see nothing special in the commit data (fetched via an API call) except the fact that it has two parent commits. I we create a new (ordinary) branch commit - we expect that author.date=committer.date, right? We create a new commit from scratch - as an atomic operation and both dates ought to be identical. Merge commit is created by GitHub, but the same reasoning should be applicable there.

rousskov (changes requested):

src/MergeContext.js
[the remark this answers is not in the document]
rousskov: When I create an ordinary commit, the dates are naturally the same because no good/reliable information is available about the changes I authored. Git could look at file timestamps instead of using current time, I guess. That would be natural too, but probably not worth the hassle in the default/common use case.

When I create a merge commit, then my expectations would depend on the type of a merge:

* If I am creating a true merge commit (which is what GitHub appears to be doing based on your "two parents" observations), then, yes, the dates would naturally be the same because the commit object represents the decision to merge (the changes have their own metadata in one of the parent chains).
* I did not know that GitHub merge commit is always a true merge commit. For example, I recalled banning merge commits in GitHub configuration some time ago and expected that to result in squashed merge commits (single parent). With squash merges, the commit author date should not be that of the commit creation date IMO (but GitHub may disagree, of course).

src/MergeContext.js
```diff
@@ -954,12 +962,50 @@ class PullRequest {
...
+        assert(!this._stagedSha());
+
+        const allEvents = await GH.getIssueEvents(this._prNumber());
+        // staging events are PR events where the bot user created a commit referencing this PR
+        let stagingEvents = allEvents.filter(ev => ev.event === "referenced" && ev.actor.login === Config.githubUserLogin());
+        if (!stagingEvents.length)
+            return;
+
+        // just in case: events should be already in chronological order already
+        stagingEvents = stagingEvents.sort((ev1, ev2) => Date.parse(ev1.created_at) - Date.parse(ev2.created_at));
```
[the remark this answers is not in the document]
eduard-bagdasaryan: >  GitHub guarantees chronological order

That is how it works, but I could not find this guarantee in the documentation.

> we may be reordering same-second events by sorting them

Do you think that sorting of equal items reorders them? I would not expect changes in case...

src/MergeContext.js
eduard-bagdasaryan: > Remove the _abandonedStagedStatuses data member

This field serves 2 purposes:
* avoiding "vain" commit recreation (covered in this thread)
* re-applying M-failed-staging-checks (after removing it at the beginning as "temporary")

So we cannot simply drop this field - at least we need to store some indication (flag) instead (and check it in stagingFailed()). Also we need to set this field early (i.e., before _checkStagingPreconditions() from where your  _checkStagingInVain() should be called) because we need to re-apply M-failed-staging-checks regardless of any other failing checks in _checkStagingPreconditions() coming prior to our call.

src/MergeContext.js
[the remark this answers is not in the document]
rousskov: > Do you think that sorting of equal items reorders them?

In many cases, it does. The question is whether this particular sort() is stable.

> BTW, it is [documented](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort) that the original order remains unchanged if elements are equal.

... starting with ES10. We are relying on ES6 support IIRC, but it is possible that the specific nodejs implementation we are using today is compliant with that specific ES10 requirement. Unfortunately, https://node.green/ does not detail/check this part AFAICT -- it only checks whether Array.sort() is available. An SO [comment](https://stackoverflow.com/questions/56063449/sorting-in-nodejs-not-stable-implications-for-memory-usage#comment98767764_56063449) says "[stable sort] was implemented first in Node 11. sort() is not stable in Node 10 or below".

What nodejs version are we using?

Can we assert somehow that it's sort() is stable or that it is compliant with ES10?

src/MergeContext.js
rousskov: > we cannot simply drop this field - at least we need to store some indication (flag) instead

Understood. I suspected something like that -- that is why I added "(if possible)". If a simple flag can be stored instead, I would try doing that to avoid the complexity of indicating what "abandonedStagedCommitStatuses" (and their failure) really mean in other contexts. Clearly, the very concept of abandonment is rather convoluted here! The function can still compute abandonedStagedCommitStatuses locally as needed, similar to my sketch, but only store a simpler well-defined flag (based on abandonedStagedCommitStatuses.failed() and other function-local decisions).

src/MergeContext.js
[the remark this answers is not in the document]
eduard-bagdasaryan: Our nodejs version is 10.19. It [seems](https://github.com/nodejs/node/issues/29446) that sort stability starts from 11 version as you noted. We can update to 11 (and assert this requirement), but I could not find how to check sort() stability independently (or its compliancy with  ES10).

src/MergeContext.js
eduard-bagdasaryan: I followed your refactoring sketch at fe1de92.

src/MergeContext.js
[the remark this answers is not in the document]
rousskov: I suggest asserting that the events are sorted instead of sorting them, with a comment explaining that we should not sort because sort() is unstable until ES10.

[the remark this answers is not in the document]
eduard-bagdasaryan: Done (7bf03bc).

src/MergeContext.js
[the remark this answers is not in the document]
rousskov: > Eduard: If we create a new (ordinary) branch commit - we expect that author.date=committer.date, right?

FWIW, when I cherry-pick a commit, the dates differ:

```
commit d68eb83 (bag51)
AuthorDate: Tue Nov 28 09:34:46 2023 -0500
CommitDate: Thu Nov 30 16:36:15 2023 -0500

    Bug 5274: Successful tunnels logged as TCP_TUNNEL/500
...    
    Cherry-picked SQUID-932-bug5274-no-retryOrBail-for-committed-srv-closure
    commit e12d6d6
```

To me, a merge commit is more similar to a cherry-picked commit than to an ordinary branch commit.

rousskov (changes requested):
AFAICT, this PR fixes the following problem: A PR fails staging checks. PR branch is then updated. Old staging checks are correctly marked as abandoned. New staging checks would succeed, but Anubis refuses to re-stage the PR (because old staging checks have failed).

```
"PR2059 (head: bfbdfd staged: 8fabd5 history: load) the staged commit is stale","time":"2025-05-08T15:56:48.266Z","v":0}
...
"PR2059 (head: bfbdfd staged: 8fabd5 history: load,stage,update) did not merge: staged commit tests failed","time":"2025-05-08T15:56:49.914Z","v":0}
```

If I am right, please update PR title (and possibly description) to reflect this bug fix as well. It does not seem to be related to "dropping M-failed-staging-checks prematurely".

PR #47, opened 2021-06-22
Recognize draft PRs as work-in-progress PRs

eduard-bagdasaryan:
PRs with the GitHub "Draft PR" setting are treated now as "WIP:" PRs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,9 @@ are satisfied:
   text displayed by GitHub for these so called "mergeable" PRs varies
   depending on whether the changes are approved and whether the PR
   branch is out of date with its target branch.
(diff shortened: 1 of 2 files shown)


rousskov:
If possible, please merge the old _wipPr() and the new _draftPr() boolean 
methods into one _draftPr() method returning "work-in-progress", "draft", 
or null. This way, no caller would be able to forget one of the two special 
conditions, and every caller would be incentivized to treat both conditions 
the same way (but still log/throw a distinct debugging/detail string).

eduard-bagdasaryan:
> merge the old _wipPr() and the new _draftPr() boolean methods into one

Done.

rousskov (approved):

PR #48, opened 2021-06-23
Re-stage PRs after empty PR branch commits

eduard-bagdasaryan:
After an empty PR branch commit, GitHub does not regenerate the tree
object of the PR merge commit. GitHub creates a new merge commit with
the old tree object instead. In this case, if the PR had a staged commit
already, the bot deemed it to be fresh because the staged commit and the
GitHub merge commit had the same commit tree (SHA). Users were naturally
surprised by the lack of a staged commit regeneration after a branch
change, especially since empty branch commits are usually used to
"awake" some automation, like CI stuck on an older (staged) commit.

This change was triggered by the following discussion:
https://github.com/squid-cache/squid/pull/651#issuecomment-639570539

### DIFF
--- a/src/MergeContext.js
+++ b/src/MergeContext.js
@@ -668,11 +668,15 @@ class PullRequest {
             const prMergeSha = await GH.getReference(this._mergePath());
             const prCommit = await GH.getCommit(prMergeSha);
             // whether the PR branch has not changed
-            const stagedCommitDate = new Date(this._stagedCommit.committer.date);
-            const prCommitDate = new Date(prCommit.committer.date);
-            if (stagedCommitDate > prCommitDate) {
-                this._log("the staged commit is fresh");
-                return true;
+            if (this._stagedCommit.tree.sha === prCommit.tree.sha) {
+                const stagedCommitDate = new Date(this._stagedCommit.committer.date);
+                const prCommitDate = new Date(prCommit.committer.date);
+                // check that a 'no-change' PR commit did not update the merge commit
+                // (which keeps the old tree object in this case)
+                if (stagedCommitDate >= prCommitDate) {
+                    this._log("the staged commit is fresh");
+                    return true;
+                }
             }
         }
         this._log("the staged commit is stale");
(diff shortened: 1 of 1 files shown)


src/MergeContext.js
```diff
@@ -668,7 +668,9 @@ class PullRequest {
             const prMergeSha = await GH.getReference(this._mergePath());
             const prCommit = await GH.getCommit(prMergeSha);
             // whether the PR branch has not changed
-            if (this._stagedCommit.tree.sha === prCommit.tree.sha) {
```
rousskov, line 671: I was expecting this PR to _add_ a freshness condition rather than replace the old seemingly essential one. Please restore this old check or discuss why it is no longer relevant.

src/MergeContext.js
```diff
@@ -668,7 +668,9 @@ class PullRequest {
...
+            const stagedCommitDate = new Date(this._stagedCommit.committer.date);
+            const prCommitDate = new Date(prCommit.committer.date);
+            if (stagedCommitDate > prCommitDate) {
```
rousskov, line 673: I doubt this matters in practice, but to support cases where the GitHub creates merge commits using the last PR branch commit date _and_ to avoid an implication that we do not support instant merge commit creation, can we treat same dates as fresh?

```suggestion
            if (stagedCommitDate >= prCommitDate) {
```

rousskov (changes requested):
Please also review my PR title/description edits and adjust further if needed.

src/MergeContext.js
eduard-bagdasaryan: Agreed - we need to compare SHAs anyway for reliability. In practice, though I guess that single 'date' check would work in all cases except when somebody (maliciously) creates an auto commit with a different tree object.

src/MergeContext.js
eduard-bagdasaryan: Agreed.

eduard-bagdasaryan (commented):
Done (e61f040).

rousskov (approved):
Thank you for addressing my concerns. Please feel free to squash-merge (with the updated title/description) and deploy.

PR #49, opened 2021-07-26
Do not uselessly flip M-abandoned-staging-checks

eduard-bagdasaryan:
When this label was added in 5ef56f6, we wanted it to be short-lived,
set and cleared during a single PR processing cycle. It was
(immediately) set when Anubis found a stale staged commit. And it was
always (implicitly) cleared at the end of the cycle, probably under the
assumption that a fresh staged commit for the same PR would be created
during the same cycle, starting a new set of checks.

However, that assumption was sometimes wrong: Anubis may not stage a new
commit for the same PR (e.g., because a review was requested or another
PR has been staged). In those cases, Anubis will keep finding the same
stale staged commit, adding a label, and removing the label on every
GitHub event that leads to that PR processing, creating a lot of noise.

We now set (or keep) the label only if the new commit was not staged (or
merged).

Not setting (or clearing) the label if a new commit was staged (or
merged) solves the problem. However, it also means that the label will
not be set-and-removed if the stale staged commit was replaced with a
new one during the same processing cycle. This is an experiment: If the
absence of that short-lived label proves to be problematic, we will
restore its immediate setting (but will still keep it if the new commit
was not staged or merged).

### DIFF
--- a/src/MergeContext.js
+++ b/src/MergeContext.js
@@ -300,13 +300,6 @@ class Labels
         return label;
     }
 
-    // adds a label, updating GitHub without waiting for pushToGitHub()
-    async addImmediately(name) {
-        const label = this.add(name);
-        if (label.needsAdditionToGitHub())
-            await this._addToGitHub(label);
-    }
-
     // removing a previously removed or missing label is a no-op
     remove(name) {
         const label = this._find(name);
@@ -489,7 +482,8 @@ class PullRequest {
 
         this._updated = false; // _update() has been called
 
-        this._abandoned = false; // abandonedStagingChecksLabel() has been set
+        // the user should see abandonedStagingChecksLabel()
+        this._signalAbandonmentOfStagingChecks = false;
 
         // truthy value contains a reason for disabling _pushLabelsToGitHub()
         this._labelPushBan = false;
@@ -986,8 +980,7 @@ class PullRequest {
         }
 
         if (!(await this._stagedCommitIsFresh())) {
-            await this._labels.addImmediately(Config.abandonedStagingChecksLabel());
-            this._abandoned = true;
+            this._signalAbandonmentOfStagingChecks = true;
             await this._enterBrewing();
             return;
         }
@@ -1013,8 +1006,11 @@ class PullRequest {
     }
 
     async _enterStaged(stagedStatuses) {
-        this._abandoned = false;
         this._prState = PrState.Staged();
+
+        // do not signal about the old staged commit when we have a new one
+        this._signalAbandonmentOfStagingChecks = false; // may already be false
+
         assert(this._stagedStatuses === null);
         if (stagedStatuses)
             this._stagedStatuses = stagedStatuses;
@@ -1025,6 +1021,9 @@ class PullRequest {
 
     _enterMerged() {
         this._prState = PrState.Merged();
+
+        // do not signal about the old staged commit when we have a merged one
+        this._signalAbandonmentOfStagingChecks = false; // may already be false
     }
 
     // loads raw PR metadata from GitHub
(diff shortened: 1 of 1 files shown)


src/MergeContext.js
```diff
@@ -489,6 +489,8 @@ class PullRequest {
 
         this._updated = false; // _update() has been called
 
+        this._abandoned = false; // abandonedStagingChecksLabel() has been set
```
rousskov, line 492: It is not clear why we could not ask this._labels whether they have that specific label set. If we do need this data member, then the description needs adjustment, but I cannot suggest anything specific yet because of the other code oddities in the PR diff.

src/MergeContext.js
```diff
@@ -985,6 +987,7 @@ class PullRequest {
 
         if (!(await this._stagedCommitIsFresh())) {
             await this._labels.addImmediately(Config.abandonedStagingChecksLabel());
+            this._abandoned = true;
```
rousskov, line 990: Could the label be set (from GitHub point of view) at this point already? Just a question; I am trying to understand the code; not requesting any changes here.

src/MergeContext.js
```diff
@@ -1010,6 +1013,7 @@ class PullRequest {
     }
 
     async _enterStaged(stagedStatuses) {
+        this._abandoned = false;
```
rousskov, line 1016: This reset looks odd because it lacks the code to remove the previously set label (if any). Why do we not need to remove the label here?

src/MergeContext.js
```diff
@@ -1311,6 +1315,9 @@ class PullRequest {
             else
                 this._removePositiveStagingLabels();
 
+            if (this._prState && this._prState.brewing() && this._abandoned)
+                this._labels.add(Config.abandonedStagingChecksLabel());
```
rousskov, line 1319: If `this._abandoned` is true, then we have already set the label by calling this._labels.addImmediately(). Why do we need to add it again?

rousskov (commented):
I have trouble understanding the diff. Please see inlined changed requests for specific questions.

rousskov (approved):
Please review my untested commits (up to aed3c46) that implement an alternative solution.

src/MergeContext.js
```diff
@@ -984,7 +980,7 @@ class PullRequest {
         }
 
         if (!(await this._stagedCommitIsFresh())) {
-            await this._labels.addImmediately(Config.abandonedStagingChecksLabel());
+            this._signalAbandonmentOfStagingChecks = true;
```
eduard-bagdasaryan, line 983: This change cancels the idea from 5ef56f6, i.e., to reflect the 'abandoning' fact in the PR history. Now there will be no abandonedStagingChecksLabel() if a staged commit becomes stale (and abandoned) and another staged commit is immediately created. Is this the intention? I could not get this from the [detailed] 3394c6c description.

rousskov: > Now there will be no abandonedStagingChecksLabel() if a staged commit becomes stale (and abandoned) and another staged commit is immediately created.

Agreed. I forgot that we apparently wanted that signal. On one hand, it feels like the creation of another staged commit is enough of a clue on its own, but perhaps that clue gets obscured in the other noise somehow. I suggest to get rid of that signal as an experiment. If we realize it was actually useful, then we will restore it (hopefully without recreating the bug this PR is fixing). AFAICT, the difficult parts of this PR will still be needed even if we decide to resurrect that signal later.

rousskov: I adjusted the PR description to reflect the current code and the above suggestion, but we can still refactor this in this PR if we are sure that short-lived signal should be kept.

eduard-bagdasaryan: > get rid of that signal as an experiment.

Ok, let's try to keep things simple for now. I am fine with this PR and we can merge it then.

PR #50, opened 2021-09-27
Added the core_developers configuration option

eduard-bagdasaryan:
Recently we started to experience persistent problems with
getCollaborators() GitHub API call, which incorrectly returned
less users than there really were. Now we specify core
collaborators manually, via a configuration option.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,8 +20,8 @@ below.
   confuse this general/imprecise term with the atomic action of merging
   one git branch into another.
 
-* _Core developer_: From the bot point of view, a core developer is a
-  GitHub user with write access to the configured repository (`config::repo`).
+* _Core developer_: A GitHub user listed in the config::core_developers
+  configuration parameter.
 
 
 ## Which pull requests are eligible for merging?
@@ -296,7 +296,7 @@ All configuration fields are required.
 *staging_branch* | The name of the bot-maintained git branch used for testing PR changes as if they were merged into their target branch. | auto
 *necessary_approvals* | The minimal number of core developers required for a PR to be merged. PRs with fewer votes are not merged, regardless of their age. | 1
 *sufficient_approvals* | The minimal number of core developers required for a PR to be merged fast (i.e., without waiting for `config::voting_delay_max`) | 2
-*core_developers* | The comma-separated list of login=id pairs representing GitHub core developers. | ""
+*core_developers* | The comma-separated list of login=id pairs, specifying GitHub login and ID for core developers. | ""
 *voting_delay_min*| The minimum merging age of a PR. Younger PRs are not merged, regardless of the number of votes. The PR age string should comply with [timestring](https://github.com/mike182uk/timestring) parser. | "2d"
 *voting_delay_max* | The maximum merging age of a PR that has fewer than `config::sufficient_approvals` votes. The PR age string should comply with [timestring](https://github.com/mike182uk/timestring) parser. | "10d"
 *staging_checks*| The expected number of CI tests executed against the staging branch. | 2
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ below.
   confuse this general/imprecise term with the atomic action of merging
   one git branch into another.
 
-* _Core developer_: A GitHub user listed in the config::core_developers
+* _Core developer_: A GitHub user listed in the `config::core_developers`
   configuration parameter.
 
 
--- a/README.md
+++ b/README.md
@@ -296,7 +296,7 @@ All configuration fields are required.
 *staging_branch* | The name of the bot-maintained git branch used for testing PR changes as if they were merged into their target branch. | auto
 *necessary_approvals* | The minimal number of core developers required for a PR to be merged. PRs with fewer votes are not merged, regardless of their age. | 1
 *sufficient_approvals* | The minimal number of core developers required for a PR to be merged fast (i.e., without waiting for `config::voting_delay_max`) | 2
-*core_developers* | The comma-separated list of GitHub account IDs representing core developers (collaborators with push access rights) | ""
+*core_developers* | The comma-separated list of login=id pairs representing GitHub core developers. | ""
 *voting_delay_min*| The minimum merging age of a PR. Younger PRs are not merged, regardless of the number of votes. The PR age string should comply with [timestring](https://github.com/mike182uk/timestring) parser. | "2d"
 *voting_delay_max* | The maximum merging age of a PR that has fewer than `config::sufficient_approvals` votes. The PR age string should comply with [timestring](https://github.com/mike182uk/timestring) parser. | "10d"
 *staging_checks*| The expected number of CI tests executed against the staging branch. | 2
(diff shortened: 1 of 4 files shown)


src/Config.js
```diff
@@ -43,6 +44,16 @@ class ConfigOptions {
...
+
+        let developerPairs = this._coreDevelopers.split(',').map(i => i.trim());
+        this._coreDeveloperIds = new Map();
+        for (let pair of developerPairs) {
+            let p = pair.split('=').map(i => i.trim());
+            assert(p.length === 2);
+            assert(!this._coreDeveloperIds.has(p[0]));
+            this._coreDeveloperIds.set(p[0], Number(p[1]));
+        }
+        assert(this._sufficientApprovals <= this._coreDeveloperIds.size);
```
rousskov, line 56: If possible, let's check both:

```suggestion
        // check that it is actually possible to get all the configured votes
        assert(this._sufficientApprovals <= this._coreDeveloperIds.size);
        assert(this._necessaryApprovals <= this._coreDeveloperIds.size);
```

src/MergeContext.js
```diff
@@ -501,13 +501,15 @@ class PullRequest {
...
-        for (let collaborator of pushCollaborators) {
-            if (requestedReviewers.includes(collaborator.login)) {
-                this._log("requested core reviewer: " + collaborator.login);
+        for (let reviewerPair of requestedReviewers.entries()) {
+            const reviewer = reviewerPair[0];
+            const requestedReviewerIsCore = Config.coreDeveloperIds().get(reviewer);
+            if (requestedReviewerIsCore) {
+                const reviewerId = reviewerPair[1];
+                assert(requestedReviewerIsCore === reviewerId);
+                this._log("requested core reviewer: " + reviewer);
```
rousskov, line 512: Please move this debugging up, _before_ the assert(). If we can print reviewerPair instead (and get both values), we should.

src/Config.js
```diff
@@ -82,6 +93,8 @@ class ConfigOptions {
     votingDelayMin() { return this._votingDelayMin; }
     stagingChecks() { return this._stagingChecks; }
     loggerParams() { return this._loggerParams; }
+    // returns a Map of (login,id) pairs
+    coreDeveloperIds() { return this._coreDeveloperIds; }
```
rousskov, line 97: AFAICT, nobody really wants this method right now. What the callers want is something like this:

```JS
// whether the given GitHub user is a core developer
coreDeveloper(role, userLogin, userId)
{
    const coreId = Config.coreDeveloperIds().get(userLogin);
    if (!coreId)
        return false;
    this._log(`found core ${role}: ${userLogin}={userId}`);
    // die if we are misconfigured and/or the userLogin has moved to another user
    assert.strictEqual(coreId, userId);
    return true;
}
```

src/Config.js
```diff
@@ -43,6 +44,16 @@ class ConfigOptions {
...
             assert(v !== undefined );
         }
+
+        let developerPairs = this._coreDevelopers.split(',').map(i => i.trim());
+        this._coreDeveloperIds = new Map();
+        for (let pair of developerPairs) {
+            let p = pair.split('=').map(i => i.trim());
+            assert(p.length === 2);
+            assert(!this._coreDeveloperIds.has(p[0]));
+            this._coreDeveloperIds.set(p[0], Number(p[1]));
```
rousskov, line 54: Please assert that the ID string parsed successfully _and_ that it is _not_ zero. The rest of the code depends on that non-falsy assumption AFAICT.

README.md
```diff
@@ -296,6 +296,7 @@ All configuration fields are required.
 *staging_branch* | The name of the bot-maintained git branch used for testing PR changes as if they were merged into their target branch. | auto
 *necessary_approvals* | The minimal number of core developers required for a PR to be merged. PRs with fewer votes are not merged, regardless of their age. | 1
 *sufficient_approvals* | The minimal number of core developers required for a PR to be merged fast (i.e., without waiting for `config::voting_delay_max`) | 2
+*core_developers* | The comma-separated list of login=id pairs representing GitHub core developers. | ""
```
rousskov, line 299: Please do not forget to also adjust the `From the bot point of view, a core developer is...` paragraph in the beginning of README.md. I suggest: `From the bot point of view, a core developer is a GitHub user listed in the `config::core_developers` configuration parameter.`

src/Config.js
```diff
@@ -43,6 +44,16 @@ class ConfigOptions {
         for (let v of allOptions) {
             assert(v !== undefined );
         }
+
+        let developerPairs = this._coreDevelopers.split(',').map(i => i.trim());
+        this._coreDeveloperIds = new Map();
```
rousskov, line 49: If possible, I would reorder to better "announce" what we are after and make the temporary/helper list constant:

```suggestion
        this._coreDeveloperIds = new Map();
        const developerPairs = this._coreDevelopers.split(',').map(i => i.trim());
```

src/Config.js
```diff
@@ -32,6 +32,7 @@ class ConfigOptions {
         this._stagingChecks = conf.staging_checks;
         this._loggerParams = conf.logger_params;
         this._approvalUrl = conf.approval_url;
+        this._coreDevelopers = conf.core_developers;
```
rousskov, line 35: Do we have to remember this "raw" value? If we can just use the derived this._coreDeveloperIds map, let's do that.

rousskov (changes requested):
Just some polishing, but one of the change requests may reduce the amount of new code duplication quite a bit! Please adjust further as needed -- I was typing those suggestions without testing.

src/Config.js
eduard-bagdasaryan: Done.

src/MergeContext.js
eduard-bagdasaryan: I removed this line because the new _coreDeveloper() does this debugging.

src/Config.js
eduard-bagdasaryan: Done.

eduard-bagdasaryan (commented):
Done (3e0ad82).

src/MergeContext.js
```diff
@@ -497,19 +497,26 @@ class PullRequest {
         return null;
     }
 
+    // whether the given GitHub user is a core developer
+    _coreDeveloper(role, userLogin, userId) {
+        const coreId = Config.coreDeveloperIds().get(userLogin);
+        if (!coreId)
+            return false;
+        this._log(`found core ${role}: ${userLogin}={userId}`);
```
rousskov, line 505: Just polishing to better blend with the actual `role` parameter values:

```suggestion
        this._log(`${role} is a core developer: ${userLogin}={userId}`);
```

rousskov (approved):
I think there was a misunderstanding regarding ID parsing. The rest is polishing. If you are happy with the code after these final corrections, please feel free to merge and deploy.

eduard-bagdasaryan (commented):
Done at e617aca.

src/MergeContext.js
rousskov: Somehow the old line did not go away when the replacement was added. We can fix it later, I guess.

PR #51, opened 2021-10-05
Do not assume that the 'PR approval' status check is always required

eduard-bagdasaryan:
StatusChecks methods did not expect that 'PR approval' could be
optional. This caused assertions because 'PR approval' _is_ optional for
some Squid branches (other from master).

### DIFF
--- a/src/MergeContext.js
+++ b/src/MergeContext.js
@@ -156,7 +156,8 @@ class StatusChecks
     //   for staged commits: the bot-configured number of required checks (Config.stagingChecks()),
     //   for PR commits: GitHub configured number of required checks (requested from GitHub)
     // context: either "PR" or "Staging";
-    constructor(expectedStatusCount, context) {
+    // approvalRequired: whether the Config.approvalContext() status check is required
+    constructor(expectedStatusCount, context, approvalRequired) {
         assert(expectedStatusCount !== undefined);
         assert(expectedStatusCount !== null);
         assert(context);
@@ -164,6 +165,7 @@ class StatusChecks
         this.context = context;
         this.requiredStatuses = [];
         this.optionalStatuses = [];
+        this.statusesForApproval = approvalRequired ? this.requiredStatuses : this.optionalStatuses;
     }
 
     addRequiredStatus(requiredStatus) {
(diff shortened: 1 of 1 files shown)


src/MergeContext.js
```diff
@@ -628,7 +630,8 @@ class PullRequest {
     async _getPrStatuses() {
         const requiredContexts = await this._getRequiredContexts();
         const combinedPrStatus = await GH.getStatuses(this._prHeadSha());
-        let statusChecks = new StatusChecks(requiredContexts.length, "PR");
+        const approvalContextRequired = requiredContexts.some(el => el.trim() === Config.approvalContext());
+        let statusChecks = new StatusChecks(requiredContexts.length, "PR", approvalContextRequired);
```
rousskov, line 634: The new parameter looks complex, dangerous, and hopefully unnecessary to me:
* complex: The caller cannot know the parameter value -- it has to compute it, and the computation is not trivial.
* dangerous: The caller can easily forget to supply the value. For example, the second caller does not supply it, raising "forgotten or intentionally omitted?" questions: https://github.com/measurement-factory/anubis/blob/10346426563aad5d08ede59177dee5c94418cb70/src/MergeContext.js#L651
* hopefully unnecessary: I hope we can just watch the status checks being added and cache the approvalContext object when it is added (optional or required, does not matter AFAICT).

If addRequiredStatus() and addOptionalStatus() are the only ways to add status checks to a StatusChecks object, please see if you can adjust those methods to call a new private StatusChecks::_maybeCacheApproval() method that will cache the approval status check object (after it is added to one of the containers). If possible, _maybeCacheApproval() should assert that at most one approval status check gets cached. hasApprovalStatus() will then use the cached value (if any) to compute the result.

rousskov (changes requested):

src/MergeContext.js
rousskov: Hm... Are the sets of required contexts always the same for the PR and the staged commit? If we always pass the result of this._getRequiredContexts() as a third StatusChecks constructor parameter, then we are saying that the required contexts stay the same. If that is a lie, then we would probably have to compute required/optional outside the constructor and assert that the constructor is given the result of that computation (as a third parameter).

rousskov (approved):
LGTM. Thank you for polishing this.

PR #53, opened 2022-04-25
Convert Authored-By statement into commit author

eduard-bagdasaryan:
If the PR description starts with "Authored-By:" attribute, the
specified author credentials are used as the commit author instead of
the ones provided by GitHub (i.e., the PR author). The entire
attribute line (including the trailing empty lines) is removed from
the PR message.

If PR author credentials cannot be parsed, the PR is rejected as having
an invalid description. The PR is rejected also if more than one
'Authored-by' attribute is specified.

Similar checks are applied also to 'Co-authored-by' attributes (coming
in the trailer part of the PR message). The credentials part of
this attribute follows the same syntax, but the trailer may have more
than one 'Co-authored-by'.

### DIFF
--- a/README.md
+++ b/README.md
@@ -238,6 +238,7 @@ PR merge, the target branch commit) from the following parts:
 * `author`: The name and email from the author object of the PR merge commit
   unless overwritten by the `Authored-by` header field value from the PR
   description header.
+* `date`: The date from the author object of the PR merge commit.
 
 Note that the individual commits on the PR branch (including their messages
 and authors) are ignored. The bot is effectively performing a squash merge
--- a/README.md
+++ b/README.md
@@ -239,6 +239,7 @@ GitHub markdown to plain text. However, both texts must conform to the
 further reduces the maximum PR title length to ~65 characters. PRs violating
 these limits are labeled `M-failed-description` and are not merged.
 
+
 ## Voting and PR approvals
 
 The following events disqualify a pull request from automatic merging. A
--- a/README.md
+++ b/README.md
@@ -238,7 +238,7 @@ PR merge, the target branch commit) from the following parts:
 * `author`: The name and email from the author object of the PR merge commit
   unless overwritten by the `Authored-by` header field value from the PR
   description header.
-* `date`: The date from the author object of the PR merge commit.
+* `author.date`: The date from the author object of the PR merge commit.
 
 Note that the individual commits on the PR branch (including their messages
 and authors) are ignored. The bot is effectively performing a squash merge
--- a/README.md
+++ b/README.md
@@ -196,35 +196,45 @@ GitHub markdown to plain text. However, both texts must conform to the
 further reduces the maximum PR title length to ~65 characters. PRs violating
 these limits are labeled `M-failed-description` and are not merged.
 
+
 ### Message attributes
 
-Commit message may contain special attributes, which are recognized and
+Commit description may contain special attributes, which are recognized and
 processed by the bot. If the bot cannot parse an attribute, it marks the PR
-it `M-failed-description` label.
+with `M-failed-description` label.
 
 1.`*Authored-by*` - The commit message author.
 We need to provide this attribute when the actual PR author is different from
 the author automatically provided by GitHub (which is the author of the PR branch
-first commit). The attribute should be specified at the beginning of the PR message.
-In order to avoid typos, lines starting with /\s*\S*Authored-By/ (except
-the first line and message trailers) are considered as invalid.
+first commit). The attribute should be specified at the beginning of the PR
+description and separated by an empty line from the rest of the message.  In order
+to avoid typos, the main part of the description (that is all lines excluding the
+first line and the message trailer) is subjected to a typo check, whereby lines
+starting with /\s*\S*Authored-By/ are considered as invalid.
 After the attribute value is parsed, the entire line (including empty lines
 below it) is removed from the message.
 The attribute is parsed according to the following ABNF rules:
-pr-author-paragraph = "Authored-By:" <SP> credentials eol empty-line+
-credentials = name+ "<" login "@" host ">"
-empty-line = eol
-eol = <CR>*<LF>
+```
+    pr-author-paragraph = "Authored-By:" <SP> credentials eol empty-line+
+    credentials = name+ "<" login "@" host ">"
+    empty-line = eol
+    eol = <CR>*<LF>
+```
 
 2. *Co-Authored-by* - The commit message co-author.
-These GitHub-recognized attribute allows to create a commit with multiple
-authors. If a message has an empty line followed by 'Co-authored-by: ',
-the bot treats this line and lines below as message trailer. Trailer
-section may contain multiple 'Co-authored-by' attrubutes, but other
-lines starting with /\s*\S*Authored-By/ are regarded as errors.
+This GitHub-recognized attribute allows to create a commit with multiple
+authors. It will be parsed by Anubis if it is specified within the last
+paragraph of the description (a.k.a. the trailer). Note that the bot recognizes
+the last paragraph as a trailer if and only if it contains attribute fields
+(i.e., `name: value` lines) and nothing else - otherwise this text block is
+considered belonging to the main part of the description and subjected to
+the typo check outlined above.
 The attribute is parsed according to the following ABNF rules:
-co-authors-paragraph = eol empty-line co-author-line+ empty-line\*
-co-author-line = "Co-Authored-By:" <SP> credentials eol
+```
+    co-authors-paragraph = eol empty-line co-author-line+ empty-line\*
+    co-author-line = "Co-Authored-By:" <SP> credentials eol
+```
+
 
 ## Voting and PR approvals
 
--- a/README.md
+++ b/README.md
@@ -184,6 +184,49 @@ All labels except `M-failed-staging-checks`, `M-failed-staging-other`,
 them useful when determining the current state of a PR.
 
 
+## Commit metadata in PR description
+
+Pull request description may contain a header and/or a trailer paragraphs with
+special `name: value` metadata fields documented below. Header fields are
+recognized only by the bot. Some trailer fields are recognized by the bot,
+GitHub, and/or git. When assembling the commit message, Anubis strips the
+header but keeps the trailer. Pull requests with descriptions containing
+invalid metadata are labeled `M-failed-description` and are not merged.
+
+```
+Authored-by: Actual Author <user@host>
+
+Regular PR description paragraph(s), if any, go here...
+
+Co-authored-by: Co-Author One <user1@host1>
+Co-authored-by: Co-Author Two <user2@host2>
+```
+
+* `Authored-by: Author Name <user@host>`: A custom commit author. Add this
+  field to the PR description header when Anubis should not use the pull
+  request author as the commit author. Mentioning this (or similarly spelled)
+  fields elsewhere in the PR description should trigger an
+  `M-failed-description` violation. At most one `Authored-by` field is
+  permitted. Anubis removes the header paragraph when assembling the commit
+  message because neither GitHub nor git recognize this field and because the
+  information will be preserved as the commit author field.
+
+* `Co-authored-by: Another Author Name <other@host>`: A commit co-author. This
+  field documents additional commit authors. It is recognized by
+  [GitHub](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors)
+  and, to an extent, [git](https://git-scm.com/docs/git-interpret-trailers).
+  Multiple unique fields are permitted.
+
+* Other trailer fields: The PR description trailer may contain fields that are
+  not recognized by Anubis. To avoid typos, the bot prohibits trailer fields
+  matching `/\S*Authored-by/i` except for the `Co-authored-by` field
+  documented above.
+
+Field names are case-insensitive, but the variants shown above are recommended
+for consistency sake. Anubis preserves trailer fields original spelling and
+formatting but strips any trailing whitespace.
+
+
 ## Commit message
 
 The staging commit message and, hence, the target branch commit is
@@ -195,46 +238,6 @@ GitHub markdown to plain text. However, both texts must conform to the
 72 characters/line limit. The automatically added ` #(NNN)` title suffix
 further reduces the maximum PR title length to ~65 characters. PRs violating
 these limits are labeled `M-failed-description` and are not merged.
-
-
-### Message attributes
-
-Commit description may contain special attributes, which are recognized and
-processed by the bot. If the bot cannot parse an attribute, it marks the PR
-with `M-failed-description` label.
-
-1.`*Authored-by*` - The commit message author.
-We need to provide this attribute when the actual PR author is different from
-the author automatically provided by GitHub (which is the author of the PR branch
-first commit). The attribute should be specified at the beginning of the PR
-description and separated by an empty line from the rest of the message.  In order
-to avoid typos, the main part of the description (that is all lines excluding the
-first line and the message trailer) is subjected to a typo check, whereby lines
-starting with /\s*\S*Authored-By/ are considered as invalid.
-After the attribute value is parsed, the entire line (including empty lines
-below it) is removed from the message.
-The attribute is parsed according to the following ABNF rules:
-```
-    pr-author-paragraph = "Authored-By:" <SP> credentials eol empty-line+
-    credentials = name+ "<" login "@" host ">"
-    empty-line = eol
-    eol = <CR>*<LF>
-```
-
-2. *Co-Authored-by* - The commit message co-author.
-This GitHub-recognized attribute allows to create a commit with multiple
-authors. It will be parsed by Anubis if it is specified within the last
-paragraph of the description (a.k.a. the trailer). Note that the bot recognizes
-the last paragraph as a trailer if and only if it contains attribute fields
-(i.e., `name: value` lines) and nothing else - otherwise this text block is
-considered belonging to the main part of the description and subjected to
-the typo check outlined above.
-The attribute is parsed according to the following ABNF rules:
-```
-    co-authors-paragraph = eol empty-line co-author-line+ empty-line\*
-    co-author-line = "Co-Authored-By:" <SP> credentials eol
-```
-
 
 ## Voting and PR approvals
 
--- a/README.md
+++ b/README.md
@@ -63,19 +63,9 @@ request (in the ascending PR number order), but merging may still fail.
 Each eligible pull request is processed according to the following
 algorithm:
 
-1. [Create](https://developer.github.com/v3/git/commits/#create-a-commit
-   ) a new "staging commit" with the following components:
-   * `tree`: The git tree object of the PR
-     [merge commit](https://developer.github.com/v3/pulls/#get-a-single-pull-request)
-     created by GitHub.
-   * `parents`: The HEAD commit of the PR base branch.
-   * `message`: PR title (with an appended PR number) plus the PR
-     description.
-   * `author`: The PR author object (containing login and email) of the
-     PR merge commit.
-   Note that the individual commits (including their messages) on the PR
-   branch are ignored. The bot is effectively performing a squash merge
-   using a GitHub-editable commit message.
+1. [Create](https://developer.github.com/v3/git/commits/#create-a-commit) a
+   new "staging commit" with the components described in "Staging commit"
+   section further below.
 2. Reset the staging (a.k.a. "auto") branch to the PR staging commit.
    The previous state of the staging branch is ignored.
 3. Wait for GitHub to report exactly `config::staging_checks` CI test
@@ -184,14 +174,10 @@ All labels except `M-failed-staging-checks`, `M-failed-staging-other`,
 them useful when determining the current state of a PR.
 
 
-## Commit metadata in PR description
+## PR description
 
-Pull request description may contain a header and/or a trailer paragraphs with
-special `name: value` metadata fields documented below. Header fields are
-recognized only by the bot. Some trailer fields are recognized by the bot,
-GitHub, and/or git. When assembling the commit message, Anubis strips the
-header but keeps the trailer. Pull requests with descriptions containing
-invalid metadata are labeled `M-failed-description` and are not merged.
+Pull request description may have up to three kinds of paragraphs: a header,
+regular paragraphs, and a trailer:
 
 ```
 Authored-by: Actual Author <user@host>
@@ -201,6 +187,18 @@ Regular PR description paragraph(s), if any, go here...
 Co-authored-by: Co-Author One <user1@host1>
 Co-authored-by: Co-Author Two <user2@host2>
 ```
+
+All parts are optional. Header and trailer parts are separated from the
+regular PR description paragraph(s) (or each other) by an empty line.
+Additional PR description formatting requirements are documented in the
+"Commit message" section further below.
+
+A header and trailer paragraphs consist of special `name: value` metadata
+fields documented below. Header fields are recognized only by the bot. Some
+trailer fields are recognized by the bot, GitHub, and/or git. When assembling
+the commit message, Anubis strips the header but keeps the trailer. Pull
+requests with descriptions containing invalid metadata are labeled
+`M-failed-description` and are not merged.
 
 * `Authored-by: Author Name <user@host>`: A custom commit author. Add this
   field to the PR description header when Anubis should not use the pull
@@ -227,11 +225,31 @@ for consistency sake. Anubis preserves trailer fields original spelling and
 formatting but strips any trailing whitespace.
 
 
-## Commit message
+## Staging commit
 
-The staging commit message and, hence, the target branch commit is
-formed by concatenating the PR title (with an appended PR number), an
-empty line and the PR description.
+Anubis assembles the staging git commit (and, hence, in case of a successful
+PR merge, the target branch commit) from the following parts:
+
+* `tree`: The git tree object of the PR [merge
+  commit](https://developer.github.com/v3/pulls/#get-a-single-pull-request)
+  created by GitHub.
+* `parents`: The HEAD commit of the PR base branch.
+* `message`: See the "Commit message" subsection for details.
+* `author`: The name and email from the author object of the PR merge commit
+  unless overwritten by the `Authored-by` header field value from the PR
+  description header.
+
+Note that the individual commits on the PR branch (including their messages
+and authors) are ignored. The bot is effectively performing a squash merge
+using PR description for commit message and optional metadata.
+
+
+### Commit message
+
+The staging commit message is formed by concatenating the PR title (with a PR
+number appended) and the PR description (without the header, if any) delimited
+by an empty line. Empty and header-only PR descriptions are allowed and result
+in a title-only commit message.
 
 Neither the title nor the description are currently processed to convert
 GitHub markdown to plain text. However, both texts must conform to the
--- a/README.md
+++ b/README.md
@@ -150,10 +150,10 @@ request state:
   determine what happened.  The bot does not attempt to merge this PR again
   until a human decides that this problem is resolved and removes the label
   manually.
-* `M-failed-description`: The PR title and/or description is invalid
-  (see below for PR commit message rules). The bot removes this label
-  when it revisits the PR and notices that the commit message components
-  were fixed.
+* `M-failed-description`: The PR title and/or description is invalid (see
+  below for PR description and commit message rules). The bot removes this
+  label when it revisits the PR and notices that the future commit message
+  components were fixed.
 * `M-failed-other`: All other fatal PR-specific errors. It is probably
   necessary to consult CI logs to determine what happened. Anubis will process
   the labeled PR again, optimistically assuming that the problems are
--- a/README.md
+++ b/README.md
@@ -196,6 +196,35 @@ GitHub markdown to plain text. However, both texts must conform to the
 further reduces the maximum PR title length to ~65 characters. PRs violating
 these limits are labeled `M-failed-description` and are not merged.
 
+### Message attributes
+
+Commit message may contain special attributes, which are recognized and
+processed by the bot. If the bot cannot parse an attribute, it marks the PR
+it `M-failed-description` label.
+
+1.`*Authored-by*` - The commit message author.
+We need to provide this attribute when the actual PR author is different from
+the author automatically provided by GitHub (which is the author of the PR branch
+first commit). The attribute should be specified at the beginning of the PR message.
+In order to avoid typos, lines starting with /\s*\S*Authored-By/ (except
+the first line and message trailers) are considered as invalid.
+After the attribute value is parsed, the entire line (including empty lines
+below it) is removed from the message.
+The attribute is parsed according to the following ABNF rules:
+pr-author-paragraph = "Authored-By:" <SP> credentials eol empty-line+
+credentials = name+ "<" login "@" host ">"
+empty-line = eol
+eol = <CR>*<LF>
+
+2. *Co-Authored-by* - The commit message co-author.
+These GitHub-recognized attribute allows to create a commit with multiple
+authors. If a message has an empty line followed by 'Co-authored-by: ',
+the bot treats this line and lines below as message trailer. Trailer
+section may contain multiple 'Co-authored-by' attrubutes, but other
+lines starting with /\s*\S*Authored-By/ are regarded as errors.
+The attribute is parsed according to the following ABNF rules:
+co-authors-paragraph = eol empty-line co-author-line+ empty-line\*
+co-author-line = "Co-Authored-By:" <SP> credentials eol
 
 ## Voting and PR approvals
 
--- a/src/MergeContext.js
+++ b/src/MergeContext.js
@@ -520,6 +520,9 @@ class CommitMessage
         assert(defaultAuthor);
         this._defaultAuthor = defaultAuthor;
 
+        // last paragraphs started with these keywords are not trailers
+        this.falseTrailers = ['typo', 'np', 'error', 'todo', 'motivation', 'note'];
+
         this._checkRaw(this._title);
 
         if (rawPr.body !== undefined && rawPr.body !== null)
@@ -584,11 +587,10 @@ class CommitMessage
         const prDescriptionWithoutHeader = this._extractHeader(prDescription);
 
         let prDescriptionWithoutHeaderAndTrailer = prDescriptionWithoutHeader; // may be adjusted below
-        try {
-            prDescriptionWithoutHeaderAndTrailer = this._extractTrailer(prDescriptionWithoutHeader);
-        } catch (e) {
-            // TODO: supply debugString() prefix
-            Log.Logger.info("assuming no trailer: " + e.message);
+        const trailerIndex = this._locateTrailer(prDescriptionWithoutHeader);
+        if (trailerIndex >= 0) {
+            this._parseTrailer(prDescriptionWithoutHeader.substring(trailerIndex));
+            prDescriptionWithoutHeaderAndTrailer = prDescriptionWithoutHeader.substring(0, trailerIndex);
         }
         this._parseBody(prDescriptionWithoutHeaderAndTrailer);
     }
@@ -608,7 +610,7 @@ class CommitMessage
 
     _extractHeader(prDescriptionRaw) {
         const prDescription = this._trim(prDescriptionRaw);
-        if (prDescription.startsWith('Authored-by:')) {
+        if (/^ {0,3}Authored-by:/.test(prDescription)) {
             let tokenizer = new FieldsTokenizer(prDescription);
             const authorField = tokenizer.nextField();
             assert(authorField);
@@ -621,16 +623,19 @@ class CommitMessage
         }
     }
 
-    _extractTrailer(prDescriptionWithoutHeaderRaw) {
-        const prDescriptionWithoutHeader = this._trim(prDescriptionWithoutHeaderRaw);
-
+    // returns the trailer index in the description or -1
+    _locateTrailer(prDescriptionWithoutHeaderRaw) {
         // index of the last occurrence of '\n\n'
-        const trailerIndex = prDescriptionWithoutHeader.search(/\n\n(?!.*\n\n)/s);
-        if (trailerIndex < 0) // the text has at most one paragraph
-            trailerIndex = 0;
+        let trailerCandidateIndex = prDescriptionWithoutHeaderRaw.search(/\n\n(?!.*\n\n)/s);
+        // Does the text have at most one paragraph?
+        trailerCandidateIndex = trailerCandidateIndex < 0 ? 0 : trailerCandidateIndex + 2
+        const trailerCandidate = prDescriptionWithoutHeaderRaw.substring(trailerCandidateIndex);
+        const trailerStart = trailerCandidate.match(/^ {0,3}(\S+): /);
+        if (trailerStart && !this.falseTrailers.includes(trailerStart[1].toLowerCase()))
+            return trailerCandidateIndex;
 
-        this._parseTrailer(prDescriptionWithoutHeader.substring(trailerIndex));
-        return prDescriptionWithoutHeader.substring(0, trailerIndex);
+        Log.Logger.info("assuming no trailer");
+        return -1;
     }
 
     _parseBody(prDescriptionWithoutHeaderAndTrailerRaw) {
--- a/src/MergeContext.js
+++ b/src/MergeContext.js
@@ -475,7 +475,7 @@ class PullRequest {
         // major methods we have called, in the call order (for debugging only)
         this._breadcrumbs = [];
 
-        this._messageValid = true;
+        this._messageValid = null;
 
         this._prState = null; // calculated PrState
 
@@ -688,7 +688,7 @@ class PullRequest {
             // check this separately because GitHub does not recreate PR merge commits
             // for conflicted PR branches (leaving stale PR merge commits).
             this._prMergeable() &&
-            await this._messageIsFresh()) {
+            this._messageIsFresh()) {
 
             const prMergeSha = await GH.getReference(this._mergePath());
             const prCommit = await GH.getCommit(prMergeSha);
@@ -794,7 +794,7 @@ class PullRequest {
 
         assert(!this._prState.merged());
 
-        this._validatePrMessage();
+        this._messageValid = this._validatePrMessage();
         this._log("messageValid: " + this._messageValid);
 
         this._approval = await this._checkApproval();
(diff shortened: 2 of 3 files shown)


src/MergeContext.js
```diff
@@ -906,6 +955,38 @@ class PullRequest {
...
+            this._warn(`Invalid '${attr}' line format`);
+            return null;
+        }
+        if (cred[0].includes(',')) {
+            this._warn(`Invalid '${attr}' line: commas are not allowed`);
+            return null;
+        }
+        const now = new Date();
+        return {name: cred[1].trim(), email: cred[2].trim(), date: now.toISOString()};
+    }
```
rousskov, line 975: I do not think we should risk this ephemeral `date` field getting into the repository. I am guessting that you wanted the structure of the returned author object to match the structure of the _stagedCommit.author, but since only one of them can contain valid dates, I think it is too dangerous to do that. The two structures should not be viewed as compatible. While _stagedCommit.author comes from GitHub, we can create a basic NameMail class to represent the Authored-by and Co-authored-by information (just name and email, no date).

This leaves the question of setting the right AuthorDate for staged commits of Authored-by PRs. I think that date should be the same as it would have been if Authored-by did not exist. In other words, adding or changing an Authored-by header should not change the AuthorDate field of the merged commit.

rousskov (changes requested):
It is difficult for me to understand parts of this code. I left a few refactoring requests that I hope will make it easier to read; please adjust my sketches as needed, of course. The rest are a few specific corrections/suggestions.

src/MergeContext.js
```diff
@@ -497,6 +497,13 @@ class PullRequest {
...
         // truthy value contains a reason for disabling _pushLabelsToGitHub()
         this._labelPushBan = false;
+
+        // the 'Authored-by' credentials from the PR message or null
+        this._authoredBy = undefined;
+
+        // PR message where the special attributes (such as 'Authored-by') were successfully parsed and removed or
+        // null on a parsing error
+        this._preprocessedPrBodyCache = undefined;
     }
```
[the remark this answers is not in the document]
eduard-bagdasaryan: I implemented CommitMessage, following your suggestion with minor differences.

src/MergeContext.js
```diff
@@ -1064,13 +1145,19 @@ class PullRequest {
...
+    _messageIsFresh() {
+        if (this._preprocessedPrMessage() !== this._stagedCommit.message)
+            return false;
+        if (this._authoredBy) {
+            const author = this._stagedCommit.author;
+            if (this._authoredBy.name !== author.name || this._authoredBy.email !== author.email)
+                return false;
+        }
+        return true;
     }
```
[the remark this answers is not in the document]
eduard-bagdasaryan: Restored.

src/MergeContext.js
eduard-bagdasaryan: I removed the date field - the date value is now provided from the merge commit (i.e., the date used with the default author). I do not think we should overcomplicate with an additional NameMail (at least for now) - just returning the unnamed hash (with two fields) should be enough.

src/MergeContext.js
```diff
@@ -906,6 +955,38 @@ class PullRequest {
...
+        const body = this._prBody();
+        if (!body.startsWith(attr))
+            return body;
+        const lineEnd = body.search(/$/m);
+        assert(lineEnd >= 0);
+        assert(this._authoredBy === undefined);
+        this._authoredBy = this._parseAuthor(attr, body.substring(attr.length, lineEnd));
+        assert(this._authoredBy !== undefined);
+        return this._authoredBy ? body.substring(lineEnd).replace(/^\s*\n+/g, '') : null;
+    }
```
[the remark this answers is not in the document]
eduard-bagdasaryan: I added FieldsTokenizer and also InvalidFieldsTokenizer (for typo checks). These helpers probably improved the code reliability (and readability) but increased its size significantly.

src/MergeContext.js
```diff
@@ -848,20 +861,56 @@ class PullRequest {
...
+            }
         }
+
+        return this._validatePrMessageAttributes();
+    }
+
+    _validatePrMessageAttributes() {
+        const origBody = this._preprocessedPrBody();
+        assert(origBody);
+        const misplacedAuthor = /^\s*\S*[aA]uthored-[bB]y/m;
```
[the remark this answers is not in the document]

eduard-bagdasaryan (commented):
Done (up to 7a863d4).

src/MergeContext.js
```diff
@@ -454,6 +454,202 @@ class BranchPosition
...
+        let lines = this._body.split('\n');
+        while (lines.length > 0) {
+            const line = lines.shift();
+            if (line.match(authorExp)) {
+                if (this._attributes.some(attr => line.startsWith(attr + ':')))
+                    continue;
+                throw new Error(`a misused '*Authored-by' attribute in the message`);
+            }
+        }
+    }
```
rousskov, line 589: Perhaps I am missing something, but I see no reason to add checkAttributes() and a fixed list of field names like this._attributes: Each loop iterating nextField()s already knows which fields are acceptable in _that_ loop context (header or trailer) and is in a much better position to handle things like unknown fields (that we want to allow in the trailer). What will change/break if we remove all the functionality associated with this._attributes? If nothing, let's remove this method and related code.

src/MergeContext.js
```diff
@@ -454,6 +454,202 @@ class BranchPosition
...
+            // index of the last occurrence of '\n\n'
+            const trailerIndex = this._body.search(/\n\n(?!.+\n\n)/s);
+            if (trailerIndex >= 0) {
+                this._trailer = this._body.substring(trailerIndex).trim();
+                this._body = this._body.substring(0, trailerIndex); // trim in _parseBody()
+                this._parseTrailer();
+            }
+
+            this._parseBody();
+
```
rousskov, line 608: The proposed parseTrailer(), parseBody/Header() order does not feel natural. If possible, please refactor using this high-level sketch:

```js
        this.extractHeader();

        try {
           this.extractTrailer();
        }
        catch (e) {
             log(... assuming no trailer: e ...);
         }

        this._checkForTypos();
 ```

where extractHeader() is something along these lines:

```js
        if (this._body.startsWith('Authored-by:')) {
            const tokenizer = new FieldsTokenizer(this._body);
            const authorFieldValue = tokenizer.nextField('Authored-by');
            this._author = this._parseAuthor(authorFieldValue);
            if (tokenizer.nextOptionalField())
                throw ... headers can only have one (Authored-by) field ...
            this._body = tokenizer.remaining();
        }
```

src/MergeContext.js
```diff
@@ -454,6 +454,202 @@ class BranchPosition
...
+        this._trailer = undefined;
+        // the (optional) Authored-by meta information extracted from the PR description header
+        this._author = undefined;
+        this._error = undefined; // Error object in a case of a parsing problem
+        // message attributes supported by the bot
+        this._attributes = ["Authored-by", "Co-authored-by"];
+
+        if (rawPr.body !== undefined && rawPr.body !== null) {
+            this._body = rawPr.body.replace(/\r+\n/g, '\n').trimEnd();
+            this._parse();
```
rousskov, line 514: I would not store raw/unparsed _body because it creates the same set of problems we are trying to solve by adding this class: The answer to the "What is a _body?" question becomes very complex.

```suggestion
            const trimmedPrDescripton = rawPr.body.replace(/\r+\n/g, '\n').trimEnd();
            this._parse(trimmedPrDescripton);
```

src/MergeContext.js
```diff
@@ -454,6 +454,202 @@ class BranchPosition
...
+        return this._lines.length === 0;
+    }
+
+    // returns the unparsed (yet) string
+    remaining() {
+        return this._lines.join('\n');
+    }
+}
+
+// Converts the raw PR message into the staged commit message.
```
rousskov, line 495: "PR message" sounds odd/undefined, and we are using many PR parts, not just the PR title/description.

```suggestion
// Computes future commit message from raw PR
```

src/MergeContext.js
```diff
@@ -755,8 +951,8 @@ class PullRequest {
         if (this._draftPr())
             throw this._exObviousFailure("just a draft");
 
-        if (!this._messageValid)
-            throw this._exLabeledFailure("invalid commit message", Config.failedDescriptionLabel());
+        if (!this._commitMessage.valid())
+            throw this._exLabeledFailure("invalid commit message: " + this._commitMessage.errorMessage(), Config.failedDescriptionLabel());
```
rousskov, line 955: If possible, let's undo this and related changes that create a notion of an "invalid this._commitMessage": When Anubis cannot create a commit message, log the problem and set this._commitMessage to null. The _commitMessage object should either be in a "valid" state or not be stored at all.

src/MergeContext.js
```diff
@@ -1064,13 +1225,30 @@ class PullRequest {
...
         this._rawPr = pr;
+        this._commitMessage = new CommitMessage(this._rawPr);
     }
 
     // Whether the commit message configuration remained intact since staging.
-    async _messageIsFresh() {
-        const result = this._prMessage() === this._stagedCommit.message;
+    _messageIsFresh() {
+        if (!this._commitMessage.valid()) {
+            this._log("staged commit message is invalid (and stale)");
```
rousskov, line 1234: It may not be stale in a direct sense of that word, AFAICT, because our validity criteria may have changed while the previously created message was staged (even though there were no PR changes at all):

```suggestion
            this._log("staged commit message became invalid (and will be treated as stale)");
```

rousskov (changes requested):
Excellent progress! This is not a comprehensive review, but I hope that this set of (mostly high-level) change requests will simplify this further.

eduard-bagdasaryan (commented):
Done (up to 3bb456f).

src/MergeContext.js
```diff
@@ -454,6 +454,201 @@ class BranchPosition
...
+        // successfully parsed
+        this._trailer = trailer;
+    }
+
+    _extractTrailer(removedHeaderDescription) {
+        // index of the last occurrence of '\n\n'
+        const trailerIndex = removedHeaderDescription.search(/\n\n(?!.+\n\n)/s);
+
+        if (trailerIndex < 0) {
+            // check a special case when the entire message may represent a trailer
```
rousskov, line 634: Please see if you can simplify (and avoid duplication) by refactoring like this:

```js
if (trailerIndex < 0) // the text has at most one paragraph
    trailerIndex = 0;

this._parseTrailer(removedHeaderDescription.substring(trailerIndex));
return removedHeaderDescription.substring(0, trailerIndex);
```

src/MergeContext.js
```diff
@@ -454,6 +454,201 @@ class BranchPosition
...
+                this._parseAuthor(field);
+            // and skip any other field
+        }
+        // successfully parsed
+        this._trailer = trailer;
+    }
+
+    _extractTrailer(removedHeaderDescription) {
+        // index of the last occurrence of '\n\n'
+        const trailerIndex = removedHeaderDescription.search(/\n\n(?!.+\n\n)/s);
```
rousskov, line 631: Nitpick: Only the following expression would actually match the `last occurrence of '\n\n'` description:
```suggestion
        const trailerIndex = removedHeaderDescription.search(/\n\n(?!.*\n\n)/s);
```

The code will probably work regardless of this change, but it is best to keep the comment and the code in sync.

src/MergeContext.js
```diff
@@ -454,6 +454,201 @@ class BranchPosition
...
+        // Author in the {name, email, date} format.
+        // The default author will be used in the future commit if the 'Authored-by' attribute is missing.
+        this._defaultAuthor = defaultAuthor;
+
+        this._checkRaw(this._title);
+
+        if (rawPr.body !== undefined && rawPr.body !== null) {
+            const trimmedPrDescription = this._trim(rawPr.body.replace(/\r+\n/g, '\n'));
+            this._checkRaw(trimmedPrDescription);
+            this._parse(trimmedPrDescription);
```
rousskov, line 518: I think each check/parse-like method should assume that it is given untrimmed input (and trim as needed). Otherwise, it is trivial to forget to trim "between" parts when splitting the raw description into parts and giving those parts to part-specific checkers/parsers...

```suggestion
            const prDescription = rawPr.body.replace(/\r+\n/g, '\n');
            this._checkRaw(prDescription);
            this._parse(prDescription);
```

.. and adjust the rest of the parsing code accordingly.

src/MergeContext.js
```diff
@@ -454,6 +454,201 @@ class BranchPosition
...
+                throw new Error(`bad character at line ${i}, offset ${invalidPosition}`);
+
+            if (line.length > 72)
+                throw new Error(`too long line '${line}'`);
+        }
+    }
+
+    // parses attributes of the raw PR description
+    _parse(trimmedPrDescription) {
+        const removedHeaderDescription = this._extractHeader(trimmedPrDescription);
```
rousskov, line 573: In line with other change requests:

```suggestion
    _parse(prDescription) {
        const prDescriptionWithoutHeader = this._extractHeader(prDescription);
```

src/MergeContext.js
```diff
@@ -454,6 +454,201 @@ class BranchPosition
...
+
+        try {
+            this._extractTrailer(removedHeaderDescription);
+        } catch (e) {
+            // TODO: supply debugString() prefix
+            Log.Logger.info("assuming no trailer: " + e.message);
+            this._body = removedHeaderDescription;
```
rousskov, line 580: Like elsewhere, this code should not trust that others have trimmed input for us:

```suggestion
            this._parseBody(prDescriptionWithoutHeader);
```

... where _parseBody() trims its parameter, checks it for typos using _checkForTypos(), and sets this._body to the checked trimmed input.

src/MergeContext.js
```diff
@@ -454,6 +454,201 @@ class BranchPosition
...
+                throw new Error(`too long line '${line}'`);
+        }
+    }
+
+    // parses attributes of the raw PR description
+    _parse(trimmedPrDescription) {
+        const removedHeaderDescription = this._extractHeader(trimmedPrDescription);
+
+        try {
+            this._extractTrailer(removedHeaderDescription);
```
rousskov, line 576: If we have two methods called extracFoo(), they should behave similarly. If the first one returns the remaining part, the second one should do the same:

```suggestion
            const prDescriptionWithoutHeaderAndTrailer = this._extractTrailer(prDescriptionWithoutHeader);
            this._parseBody(prDescriptionWithoutHeaderAndTrailer);
```

src/MergeContext.js
```diff
@@ -454,6 +454,201 @@ class BranchPosition
...
+            const field = tokenizer.nextField();
+            if (field.name === "Co-authored-by")
+                this._parseAuthor(field);
+            // and skip any other field
+        }
+        // successfully parsed
+        this._trailer = trailer;
+    }
+
+    _extractTrailer(removedHeaderDescription) {
```
rousskov, line 629:
```suggestion
    _extractTrailer(prDescriptionWithoutHeaderRaw) {
        const prDescriptionWithoutHeader = this._trim(prDescriptionWithoutHeaderRaw);
```

src/MergeContext.js
```diff
@@ -454,6 +454,201 @@ class BranchPosition
...
+            throw new Error(`an empty trailer`);
+
+        while (!tokenizer.atEnd()) {
+            const field = tokenizer.nextField();
+            if (field.name === "Co-authored-by")
+                this._parseAuthor(field);
+            // and skip any other field
+        }
+        // successfully parsed
+        this._trailer = trailer;
```
rousskov, line 626: We need to reject (descriptions containing) trailers with `Authored-by` fields, I think.

```suggestion

        this._checkForTypos(trailer);
        this._trailer = trailer;
```

src/MergeContext.js
```diff
@@ -454,6 +454,201 @@ class BranchPosition
...
+            this._body = "";
+        } else {
+            const trailer = this._trim(removedHeaderDescription.substring(trailerIndex));
+            this._parseTrailer(trailer);
+            this._body = this._trim(removedHeaderDescription.substring(0, trailerIndex));
+        }
+    }
+
+    // checks whether the message has a misused attribute
+    _checkForTypos() {
```
rousskov, line 645: This method needs a `text` parameter. It should be used for both body and trailer parts.

src/MergeContext.js
```diff
@@ -454,6 +454,201 @@ class BranchPosition
...
+        const authorExp = /^\s*\S*[aA]uthored[-_]?[bB]y/m;
+        if (this._body.search(authorExp) >= 0)
+            throw new Error(`a misused '*Authored-by' attribute in the message`);
```
rousskov, line 648:
```suggestion
            throw new Error(`suspicious '*Authored-by' attribute in the PR description`);
```

src/MergeContext.js
```diff
@@ -454,6 +454,201 @@ class BranchPosition
...
+            const trailer = this._trim(removedHeaderDescription.substring(trailerIndex));
+            this._parseTrailer(trailer);
+            this._body = this._trim(removedHeaderDescription.substring(0, trailerIndex));
+        }
+    }
+
+    // checks whether the message has a misused attribute
+    _checkForTypos() {
+        const authorExp = /^\s*\S*[aA]uthored[-_]?[bB]y/m;
+        if (this._body.search(authorExp) >= 0)
```
rousskov, line 647:
```suggestion
        const possibleAuthoredByTypoRegex = /^\s*\S*[aA]uthored[-_]?[bB]y/m;
        if (this._body.search(possibleAuthoredByTypoRegex) >= 0)
```

rousskov (changes requested):
There are a couple of potentially difficult questions. The rest is minor bug fixing and polishing. The polishing comments partially overlap and may contain some contradictions; please try to keep the result consistent despite these distractions.

eduard-bagdasaryan (commented):
Partially done (a0f0e33d), a couple of questions needs clarification.

src/MergeContext.js
```diff
@@ -454,6 +454,214 @@ class BranchPosition
...
+    }
+
+    // author is a {name, value}
+    _parseAuthor(author) {
+        const cred = author.value.match(/^ ([\w][^@<>,]*) <(\S+@\S+\.\S+)>$/);
+        if (!cred)
+            throw new Error(`invalid '${author.name}' value format`);
+
+        if (cred[0].includes(','))
+            throw new Error(`invalid '${author.name}' value: commas are not allowed`);
```
rousskov, line 595:
```suggestion
            throw new Error(`${author.name} author name with a comma: ${author.value}`);
```

src/MergeContext.js
```diff
@@ -454,6 +454,214 @@ class BranchPosition
...
+            // TODO: supply debugString() prefix
+            Log.Logger.info("assuming no trailer: " + e.message);
+            this._parseBody(prDescriptionWithoutHeader);
+        }
+
+        assert(this._body !== undefined && this._body !== null);
+    }
+
+    // author is a {name, value}
+    _parseAuthor(author) {
```
rousskov, line 589:
```suggestion
    // extracts author name and email from Authored-by and similar fields
    // returns a {name, email} author tuple
    _parseAuthor(author) {
        // author is a {name, value} tuple, where name is the field name
```

I also recommend renaming the author parameter to authorField. Otherwise, `author.name` expressions in this code confuse readers (like me) because an author also has a name.

src/MergeContext.js
```diff
@@ -454,6 +454,214 @@ class BranchPosition
...
+    }
+
+    _extractHeader(prDescriptionRaw) {
+        const prDescription = this._trim(prDescriptionRaw);
+        if (prDescription.startsWith('Authored-by:')) {
+            let tokenizer = new FieldsTokenizer(prDescription);
+            const authorField = tokenizer.nextField();
+            this._author = this._parseAuthor(authorField);
+            if (!tokenizer.atEnd())
+                throw new Error(`header must have nothing but a single 'Authored-by' attribute`);
```
rousskov, line 607:
```suggestion
                throw new Error(`unexpected header lines after a single Authored-by attribute`);
```

src/MergeContext.js
```diff
@@ -454,6 +454,214 @@ class BranchPosition
...
+            return this._trim(tokenizer.remaining());
```
rousskov, line 608: This trimming should not be our responsibility. Placing this trimming code here implies that the caller can count on us trimming the value for them, but we do not want to create that kind of dependency. Each parser should trim its raw input.

```suggestion
            return tokenizer.remaining();
```

src/MergeContext.js
```diff
@@ -454,6 +454,214 @@ class BranchPosition
...
+
+        this._trailer = trailer;
+    }
+
+    _extractTrailer(prDescriptionWithoutHeaderRaw) {
+        const prDescriptionWithoutHeader = this._trim(prDescriptionWithoutHeaderRaw);
+        // index of the last occurrence of '\n\n'
+        const trailerIndex = prDescriptionWithoutHeader.search(/\n\n(?!.*\n\n)/s);
+
+        if (trailerIndex < 0) // the text has at most one paragraph
```
rousskov, line 639: Just better line-grouping (moving an empty line up, placing it before the "index of..." comment):


```suggestion

        // index of the last occurrence of '\n\n'
        const trailerIndex = prDescriptionWithoutHeader.search(/\n\n(?!.*\n\n)/s);
        if (trailerIndex < 0) // the text has at most one paragraph
```

src/MergeContext.js
```diff
@@ -454,6 +454,214 @@ class BranchPosition
...
+            this._author = this._parseAuthor(authorField);
+            if (!tokenizer.atEnd())
+                throw new Error(`header must have nothing but a single 'Authored-by' attribute`);
+            return this._trim(tokenizer.remaining());
+        } else {
+            return prDescription;
+        }
+    }
+
+    _parseTrailer(trailerRaw) {
```
rousskov, line 614: I would move this method below the _extractTrailer() one (or even below the _parseBody() method), for a more natural order.

src/MergeContext.js
```diff
@@ -454,6 +454,214 @@ class BranchPosition
...
+            throw new Error(`an empty trailer`);
+
+        while (!tokenizer.atEnd()) {
+            const field = tokenizer.nextField();
+            if (field.name === "Co-authored-by") {
+                this._parseAuthor(field);
+                const coAuthor = JSON.stringify(this._parseAuthor(field));
+                this._log(`accepting trailer field: ${coAuthor}`);
+            }
+            // and skip any other field
```
rousskov, line 628: We are missing a _checkForTypos() call for the skipped field. We must catch and ban Authored-by lines in the trailer! It may be a good idea to add a .raw member to the nextField() result tuple so that we can pass _that_ original string to _checkForTypos() instead of reconstructing its approximation here...

src/MergeContext.js
```diff
@@ -454,6 +454,214 @@ class BranchPosition
...
+        const trailer = this._trim(trailerRaw);
+        let tokenizer = new FieldsTokenizer(trailer);
+
+        if (tokenizer.atEnd())
+            throw new Error(`an empty trailer`);
+
+        while (!tokenizer.atEnd()) {
+            const field = tokenizer.nextField();
+            if (field.name === "Co-authored-by") {
+                this._parseAuthor(field);
```
rousskov, line 624: I think the first line was accidentally left behind during logging improvements below it. We do not want to parse this field twice.

```suggestion
```

rousskov (changes requested):
Good progress, thank you, but still a few problems AFAICT. The rest is polishing.

eduard-bagdasaryan (commented):
Done (fb17776).

src/MergeContext.js
```diff
@@ -454,6 +454,214 @@ class BranchPosition
...
+            this._checkRaw(prDescription);
+            this._parse(prDescription);
+        }
+    }
+
+    // complete message for the future commit
+    whole() {
+        assert(this._author !== null && this._body !== null && this._trailer !== null);
+        let message = this._title;
+        if (this._body)
```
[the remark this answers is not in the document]
rousskov: I thought that the code was written so that this._body was always set (because this._parseBody() was called for all successfully constructed objects). However, I now see that the constructor assumes that rawPr.body may be undefined or null (and does not call _parseBody() in those cases). I missed that. Sorry.

It feels like there is a problem somewhere:

* either the lack of PR description is not normal/expected and should result in CommitMessage constructor throwing
* or undefined/null PR description just implies an empty PR description, and the CommitMessage constructor should parse that empty description instead of not parsing at all (to avoid creating a third body state that we should worry about in whole() and possibly other methods).

I suspect we should restructure the code a little so that this._body is either null or a non-empty string (but never undefined). Alternatively, we can restructure so that this._body is always a string (never null/undefined) but can be an empty string. I think the first alternative is slightly better/clearer, but I am not sure. Please pick the best alternative and implement it.

src/MergeContext.js
```diff
@@ -454,6 +454,214 @@ class BranchPosition
...
+            message += '\n\n' + this._body;
+        if (this._trailer)
```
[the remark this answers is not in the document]
rousskov: Sorry, I missed the fact that this._trailer may be left undefined. I suspect we should restructure the code a little so that this._trailer is either null or a non-empty string (but never undefined). Alternatively, we can restructure so that this._trailer is always a string (never null/undefined) but can be an empty string. I think the first alternative is slightly better/clearer, but I am not sure. Please pick and implement the same approach for _body and _trailer.

rousskov (changes requested):
Only two (similar) problems remain AFAICT, exposed by my incorrect change requests.

README.md
```diff
@@ -197,6 +197,45 @@ further reduces the maximum PR title length to ~65 characters. PRs violating
...
+### Message attributes
+
+Commit description may contain special attributes, which are recognized and
+processed by the bot. If the bot cannot parse an attribute, it marks the PR
+with `M-failed-description` label.
+
+1.`*Authored-by*` - The commit message author.
+We need to provide this attribute when the actual PR author is different from
+the author automatically provided by GitHub (which is the author of the PR branch
+first commit). The attribute should be specified at the beginning of the PR
```
rousskov, line 209: Please double check the "author of the PR branch first commit" statement. IIRC, you have [concluded](https://measurementfactory.atlassian.net/browse/SQUID-642?focusedCommentId=46179) that it is the PR author that is used as the merge commit author. This README file also says that the staged commit author is "The PR author object".

Please do not change the text even if my suspicions are correct -- I am editing this text now. Just let me know which author is used by default, first branch commit or PR. Thank you.

README.md
```diff
@@ -184,6 +184,49 @@ All labels except `M-failed-staging-checks`, `M-failed-staging-other`,
...
+  `M-failed-description` violation. At most one `Authored-by` field is
+  permitted. Anubis removes the header paragraph when assembling the commit
+  message because neither GitHub nor git recognize this field and because the
+  information will be preserved as the commit author field.
+
+* `Co-authored-by: Another Author Name <other@host>`: A commit co-author. This
+  field documents additional commit authors. It is recognized by
+  [GitHub](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors)
+  and, to an extent, [git](https://git-scm.com/docs/git-interpret-trailers).
+  Multiple unique fields are permitted.
```
rousskov, line 218: If it is not too difficult, we should reject PRs with identical trailer fields. If it is too difficult, we should drop the word "unique" from this paragraph.

rousskov:
> Alex: (but the _committer_ is).

That part is false. The committer of official commits is Anubis, not PR author. Thus, the prAuthor() cannot match when the author was customized (AFAICT). I do not think we can (easily) get the right author from the PR description that early in PR processing. Can we drop the author field from the search criteria?

README.md
eduard-bagdasaryan: >  reject PRs with identical trailer fields

Done (e7f9c7e).

eduard-bagdasaryan (commented):
Done (up to e7f9c7e).

eduard-bagdasaryan:
> Can we drop the author field from the search criteria?

I agree that we should do this because we cannot correctly determine the author of the already merged commit (somebody could have modified the custom PR author since then).

README.md
eduard-bagdasaryan: > author of the PR branch first commit

My experiment showed that the PR author is used by default (by us and by GitHub when creating merge commit). So my previous statement was wrong - "author of the PR branch first commit" has nothing to do with this. If we retrieve a PR (via API), the PR author is stored in the `user.login` field (and returned by our _prAuthor()).

src/MergeContext.js
```diff
@@ -454,6 +454,222 @@ class BranchPosition
...
+            throw new Error(`an empty trailer`);
+
+        while (!tokenizer.atEnd()) {
+            const field = tokenizer.nextField();
+            if (field.name === "Co-authored-by") {
+                const coAuthor = JSON.stringify(this._parseAuthor(field));
+                this._log(`accepting trailer field: ${coAuthor}`);
+            } else {
+                this._checkForTypos(field.raw);
+            }
```
rousskov, line 653: We should not check the `Co-authored-by` name, but let's check its value. In fact, let's check all field values (and unknown field names):

```suggestion
                this._checkForTypos(field.name);
            }
            this._checkForTypos(field.value);
```

src/MergeContext.js
```diff
@@ -454,6 +454,222 @@ class BranchPosition
...
+        }
+
+        if (trailer.length > 0)
+            this._trailer = trailer;
+    }
+
+    // checks the PR message (or its part) for some common/expected typos that may occur
+    // when filling in PR attributes
+    _checkForTypos(text) {
+        const possibleAuthoredByTypoRegex = /^\s*\S*[aA]uthored[-_]?[bB]y/m;
```
rousskov, line 663: Let's make it even more strict by making the whole pattern case-insensitive. We can relax it later if needed.

```suggestion
        const possibleAuthoredByTypoRegex = /^\s*\S*authored[-_]?by/mi;
```

src/MergeContext.js
```diff
@@ -1063,14 +1248,39 @@ class PullRequest {
...
+            } catch (e) {
+                this._logEx(e, "cannot parse commit message");
+            }
+        }
     }
 
     // Whether the commit message configuration remained intact since staging.
-    async _messageIsFresh() {
-        const result = this._prMessage() === this._stagedCommit.message;
+    _messageIsFresh() {
```
rousskov, line 1266: Since we are checking both message text and author fields, I suggest replacing "commit message configuration" with "staged commit metadata" and renaming _messageIsFresh() to _stagedCommitMetadataIsFresh().

src/MergeContext.js

src/MergeContext.js
```diff
@@ -454,6 +454,222 @@ class BranchPosition
     }
 }
 
+// Forward iterator for fields in the 'name:value' format.
+class FieldsTokenizer
+{
+    constructor(str, allowDuplicates = true) {
```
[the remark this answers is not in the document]

eduard-bagdasaryan:
> README editing. Please check 

Done.

eduard-bagdasaryan (commented):
Done with the remaining comments (up to 90fb100).

src/MergeContext.js
```diff
@@ -1063,14 +1264,39 @@ class PullRequest {
...
+        const result = this._commitMessage.whole() === this._stagedCommit.message;
         this._log("staged commit message freshness: " + result);
-        return result;
+        if (!result)
+            return false;
+
+        const oldAuthor = this._stagedCommit.author;
+        const newAuthor = this._commitMessage.author();
+        const authorIsFresh = oldAuthor.name === newAuthor.name && oldAuthor.email === newAuthor.email;
+        this._log("staged commit author freshness: " + authorIsFresh);
```
rousskov, line 1326: Why are we ignoring (i.e. not comparing) the author date here? The _commitMessage.author() gets its date from this._mergeCommit and here we are examining this._stagedCommit so it feels wrong to assume the two are the same. Or perhaps we are assuming that they may change and that change should be ignored for some reason? In that case, we should add a source code comment to explain.

src/MergeContext.js
```diff
@@ -454,6 +454,270 @@ class BranchPosition
...
+        // errors. Unfortunately, there is no reliable way to detect trailer
+        // presence. Thus, we will miss some trailers (and merge some PRs with
+        // buggy trailers), and we will validate some non-trailers (and block
+        // some valid PRs). The current implementation considers excessive
+        // blocking to be the lesser of the two evils: We treat any last
+        // paragraph that starts with /\S+:/ as a trailer _unless_ it starts
+        // with one of these common expressions.
+        this.phonyTrailers = [
+            // common paragraph labels or introductory words
+            /^TODO:/i,
```
rousskov, line 546: I documented the current approach, but I believe it is not the best approach to use here. This hard-coded list of special non-trailer labels creates long-term code maintenance overheads, and we are going to annoy PR authors when their PRs are going to be incorrectly labeled/blocked. All that is not worth a slightly improved chance that we will notice a typo in a true trailer.

Given that in-scope typos ought to be very rare and virtually all PR descriptions are reviewed by humans anyway, I think we should chose to deal with the other evil (i.e., miss more typos) and use the opposite header/trailer detection algorithm: If paragraphLabel() contains a dash (`-`), then it is a header or trailer paragraph. Otherwise, it is not. Note that none of the existing phonyTrailers contain dashes so the whole array can be removed.

rousskov (changes requested):

src/MergeContext.js
eduard-bagdasaryan: > the other evil (i.e., miss more typos) and use the opposite header/trailer detection algorithm

Ok, let's do that.

> paragraphLabel() contains a dash (-), then it is a header or trailer paragraph

Also we should take into account that some legitimate one-word trailers could go without '-'. We cannot do this accurately, however, because there is no standard for Git message trailers (each platform such as GitHub supports its own set). For example, kernel developers use [this](https://git.wiki.kernel.org/index.php/CommitMessageConventions) list of trailers, and some of them lack '-'.

rousskov: > Also we should take into account that some legitimate one-word trailers could go without '-'.

Agreed. Anubis will mistreat last paragraphs that start with a one-word field as description body. Since the probability that mistreatment resulting in an actual problem (i.e. the rejection of the misidentified body due to a `Co-authored-by` or similar field that follows that one-word field) is very low and easily worked around without Anubis changes (by reordering trailer fields to place the `Co-authored-by` field first), I think we should accept that as the (far) lesser evil.

In the future, we can always add a list of recognized/honored trailer fields (and even a special marker to start the trailer) to tighten this up further if needed.

src/MergeContext.js
eduard-bagdasaryan: We compare dates in the outer _stagedCommitIsFresh() method, just after comparing tree SHAs. I moved these checks here to have all of them in one place (c486b1). Also I the old check compared commiter.date, but it should compare author.date instead.

eduard-bagdasaryan (commented):
Done (up to c486b1d).

src/MergeContext.js
```diff
@@ -454,6 +454,252 @@ class BranchPosition
...
+        this._checkRaw(prDescription);
+        const prDescriptionWithoutHeader = this._extractHeader(prDescription);
+        const prDescriptionWithoutHeaderAndTrailer = this._extractTrailer(prDescriptionWithoutHeader);
+        this._parseBody(prDescriptionWithoutHeaderAndTrailer);
+    }
+
+    // authorField is a {name, value, raw}
+    _parseAuthor(authorField) {
+        const trimmedValue = this._trim(authorField.value);
+        const cred = trimmedValue.match(/^ ([\w][^@<>,]*) <(\S+@\S+\.\S+)>$/);
```
rousskov, line 620: The leading field value space is required for the field to be parsed. It is essentially a part of the field name:value delimiter. Let's not keep that whitespace in `.value`. Let's also trim each `.value` when parsing fields so that higher-level code like this does not have to deal with beginning and ending whitespace at all. Most of these whitespace trimming changes would go into _tokenizeAll(), where the fields are parsed, but we also need to remove the extra space from the regex here:

```suggestion
        const cred = authorField.value.match(/^([\w][^@<>,]*) <(\S+@\S+\.\S+)>$/);
```

src/MergeContext.js
```diff
@@ -454,6 +454,252 @@ class BranchPosition
...
+            return prDescription;
+        }
+    }
+
+    _extractTrailer(prDescriptionWithoutHeaderRaw) {
+        const prDescriptionWithoutHeader = this._trim(prDescriptionWithoutHeaderRaw);
+        // index of the last occurrence of '\n\n'
+        let trailerCandidateIndex = prDescriptionWithoutHeader.search(/\n\n(?!.*\n\n)/s);
+        // Does the text have at most one paragraph?
+        trailerCandidateIndex = trailerCandidateIndex < 0 ? 0 : trailerCandidateIndex + 2
```
rousskov, line 650: Missing semicolon, but I would also add parens to avoid any doubts regarding the `?` RHS expression scope.

```suggestion
        trailerCandidateIndex = trailerCandidateIndex < 0 ? 0 : (trailerCandidateIndex + 2);
```

rousskov (approved):
I left a few small polishing requests, but I hope this complicated change is ready to be (carefully) deployed after those are addressed.

src/MergeContext.js
```diff
@@ -454,6 +454,252 @@ class BranchPosition
...
+            const line = lines[i];
+            const invalidPosition = this._invalidCharacterPosition(line);
+
+            if (invalidPosition !== -1)
+                throw new Error(`bad character at line ${i}, offset ${invalidPosition}`);
+
+            // allow excessively long whitespace-only lines
+            // that some copy-pasted PR descriptions may include
+            const trimmedLine = this._trim(line);
+            if (trimmedLine.length > 72)
```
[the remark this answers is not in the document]
eduard-bagdasaryan: Ok, but why trailer? So that we could allow, e.g., long user_name+email lines for Co-authered-by - and trailers cannot be multiline?

src/MergeContext.js
eduard-bagdasaryan: Ok.

src/MergeContext.js
eduard-bagdasaryan: Ok.

src/MergeContext.js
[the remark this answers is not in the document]
rousskov: > So that we could allow, e.g., long user_name+email lines

Exactly. For example, the longest existing entry in Squid CONTRIBUTORS is 70 characters long, and that is without the header name...

> and trailers cannot be multiline?

I do not know whether they can be, but our current PR code pretty much assumes that the end of the line is the end of the field.

src/MergeContext.js
```diff
@@ -454,6 +454,254 @@ class BranchPosition
...
+        const prDescription = prDescriptionRaw.replace(/\r+\n/g, '\n');
+        this._checkRaw(prDescription);
+        const prDescriptionWithoutHeader = this._extractHeader(prDescription);
+        const prDescriptionWithoutHeaderAndTrailer = this._extractTrailer(prDescriptionWithoutHeader);
+        this._parseBody(prDescriptionWithoutHeaderAndTrailer);
+    }
+
+    // authorField is a {name, value, raw}
+    _parseAuthor(authorField) {
+        const trimmedValue = this._trim(authorField.value);
```
rousskov, line 621: Feels wrong to trim the field value (that is supposed to be trimmed for us already), especially using the special-purpose _trim() that does not trim the leading space and works with multiline text.

src/MergeContext.js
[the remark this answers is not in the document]
rousskov: BTW, I failed to add you as the author of PR 1109 because of that limit: The line was 77 characters long.

There may be an even more general use case here, accommodating long URLs and such. This is one of the reasons I would not rush to implement this enhancement in this PR.

src/MergeContext.js
```diff
@@ -454,6 +454,265 @@ class BranchPosition
...
+            throw new Error(`${authorField.name} author name with a comma: ${authorField.value}`);
+
+        return {name: cred[1].trim(), email: cred[2].trim()};
+    }
+
+    _extractHeader(prDescriptionRaw) {
+        const prDescription = this._trim(prDescriptionRaw);
+        const headerFieldName = 'Authored-by';
+        if (this._startsWithFieldName(prDescription) === headerFieldName) {
+            let tokenizer = new FieldsTokenizer(prDescription, 1);
```
rousskov, line 644: FWIW, I find the old approach -- checking atEnd() rather than supplying the expected number of fields -- cleaner/simpler. I would also expect it to require fewer changes when we add support for more header fields. Perhaps I missed something -- I did not see why that approach needed to be replaced.

src/MergeContext.js
```diff
@@ -454,6 +454,265 @@ class BranchPosition
...
+        // cannot just use trim() to preserve the first line indentation (if any).
+        return str.replace(/^\s*\n+/g, '').trimEnd();
+    }
+
+    // returns either a string containing canonical spelling of the first paragraph word
+    // that is followed by a colon, or null
+    _paragraphLabel(str) {
+        if (/^\s{4}/.test(str)) // not a regular paragraph but a quotation
+            return null;
+        const rawLabel = str.match(/^\s*(\S+)\s*:/);
```
rousskov, line 584: IIRC, we have discussed tightening this up to ban whitespace before colon:

```suggestion
        const rawLabel = str.match(/^\s*(\S+):/);
```

src/MergeContext.js
rousskov: Still feels wrong.

src/MergeContext.js
```diff
@@ -454,6 +454,265 @@ class BranchPosition
...
+            if (/^\s/.test(line))
+                throw new Error(`a field cannot start with whitespace: ${line}`);
+
+            const pos = line.search(': ');
+            if (pos < 0)
+                throw new Error(`a field without a name: value delimiter: '${line}'`);
+
+            const name = line.substring(0, pos);
+            const value = line.substring(pos+2).trim();
+            if (this._remainingFields.some(el => el.name.toUpperCase() === name.toUpperCase() &&
```
rousskov, line 495: We should probably ban field name with spaces, including trailing spaces.

eduard-bagdasaryan: Done (965d814).

src/MergeContext.js
eduard-bagdasaryan: Done (97ab1f8).

src/MergeContext.js
eduard-bagdasaryan: Done (bad37d0).

src/MergeContext.js
```diff
@@ -626,8 +629,7 @@ class CommitMessage
 
     // authorField is a {name, value, raw}
     _parseAuthor(authorField) {
-        const trimmedValue = this._trim(authorField.value);
-        const cred = trimmedValue.match(/^([\w][^@<>,]*) <(\S+@\S+\.\S+)>$/);
+        const cred = authorField.value.match(/^([\w][^@<>,]*) <(\S+@\S+\.\S+)>$/);
```
rousskov, line 632: If you have not deployed yet, I would relax this a bit to allow tabs and such:

```suggestion
        const cred = authorField.value.match(/^([\w][^@<>,]*)\s<(\S+@\S+\.\S+)>$/);
```

We trim the name value below so we only need to see one space (technically, even that single whitespace character is optional, I guess, but I am not sure we should get _that_ permissive).

eduard-bagdasaryan: Done (cd5ad03).

src/MergeContext.js
eduard-bagdasaryan: I restored the old approach at cd5ad03.

PR #54, opened 2022-08-10
Always trim trailing whitespace

eduard-bagdasaryan:
PR description and title lines are now end-trimmed - there should be no
need in these superfluous whitespaces.

### DIFF
--- a/src/MergeContext.js
+++ b/src/MergeContext.js
@@ -510,7 +510,7 @@ class CommitMessage
     constructor(rawPr, defaultAuthor) {
 
         // the (required) commit message title
-        this._title = rawPr.title + ' (#' + rawPr.number + ')';
+        this._title = this._checkAndTrimRaw(rawPr.title) + ' (#' + rawPr.number + ')';
 
         // PR description has three optional parts: [header]+[body]+[trailer]
         // The parts are separated by empty lines.
@@ -535,8 +535,6 @@ class CommitMessage
         // uses {name, email} format
         // overwrites this._defaultAuthor's name and email
         this._customAuthor = null;
-
-        this._checkRaw(this._title);
 
         if (rawPr.body !== undefined && rawPr.body !== null)
             this._parse(rawPr.body);
@@ -592,11 +590,13 @@ class CommitMessage
         return (label && label.includes('-')) ? label : null;
     }
 
-    // basic checks for an unparsed message
-    _checkRaw(message) {
+    // performs basic checks for an unparsed message
+    // trims the end of each message line and returns the result
+    _checkAndTrimRaw(message) {
         // CRs in CRLF sequences already removed
         // other CRs are not treated specially (and are banned)
         const lines = message.split('\n');
+        let trimmedLines = [];
         for (let i = 0; i < lines.length; ++i) {
             const line = lines[i];
             const invalidPosition = this._invalidCharacterPosition(line);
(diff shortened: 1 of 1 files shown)


src/MergeContext.js
```diff
@@ -510,7 +510,7 @@ class CommitMessage
     constructor(rawPr, defaultAuthor) {
 
         // the (required) commit message title
-        this._title = rawPr.title + ' (#' + rawPr.number + ')';
+        this._title = this._checkAndTrimRaw(rawPr.title) + ' (#' + rawPr.number + ')';
```
rousskov, line 513: We should remove the extra whitespace from the beginning of the title as well, even if GitHub currently does that for us. Our one-size-fits-all _checkAndTrimRaw() cannot do that (as already discussed in the other change request).

I suggest refactoring to use something like this:
* _parseRawLines() -- checks for bad characters, supports multiple lines, trims each at the end, enforces the line length limit via _checkLength()
* _parseRawWords() -- checks for bad characters, trims from both ends, does not enforce the length limit
* _checkLineLength() -- checks the length limit of a parsed or generated line (including the generated title)

src/MergeContext.js
```diff
@@ -606,17 +606,19 @@ class CommitMessage
 
             // allow excessively long whitespace-only lines
             // that some copy-pasted PR descriptions may include
-            const trimmedLine = this._trim(line);
+            trimmedLines.push(line.trimEnd());
```
rousskov, line 609: Please keep the trimmedLine variable and push it _after_ the throwing check below.

rousskov (changes requested):

src/MergeContext.js
```diff
@@ -536,8 +536,6 @@ class CommitMessage
         // overwrites this._defaultAuthor's name and email
         this._customAuthor = null;
 
-        this._checkRaw(this._title);
```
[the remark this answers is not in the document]
eduard-bagdasaryan: The old check was also a little flawed, because `ths._title` is not 'raw' but already edited title... This makes me think that our 72 character check is misplaced - this limitation should be applied to the output message (i.e., CommitMessage.whole()), not to the input/raw one. In other words, we can validate the result with certainty only after all N editing steps have been applied and the final message is formed.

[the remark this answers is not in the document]
rousskov: I do not disagree, but I did not want to bring that up in the PR about whitespace trimming. We have another Jira issue dedicated to the length check improvements... Here, we just need to avoid breaking things as far as the 72-character limit testing is concerned.

src/MergeContext.js
eduard-bagdasaryan: I followed your plan (85fe2fb) with minor differencies.

src/MergeContext.js
```diff
@@ -561,11 +566,36 @@ class CommitMessage
         return {name: this._customAuthor.name, email: this._customAuthor.email, date: this._defaultAuthor.date};
     }
 
-    // returns the position of the first non-ASCII_printable character (or -1)
-    _invalidCharacterPosition(str) {
+    // checks that the line represents only ASCII_printable characters
+    _parseRawCharacters(line) {
```
rousskov, line 570: Unlike other parsing methods, this one does not store the results:

```suggestion
    _checkRawCharacters(line) {
```

eduard-bagdasaryan: Done (04f70b3).

src/MergeContext.js
```diff
@@ -592,31 +622,9 @@ class CommitMessage
...
-            // TODO: Allow longer header (and possibly even trailer) lines.
-            if (trimmedLine.length > 72)
-                throw new Error(`too long line '${trimmedLine}'`);
-        }
-    }
-
     _parse(prDescriptionRaw) {
-        const prDescription = prDescriptionRaw.replace(/\r+\n/g, '\n');
-        this._checkRaw(prDescription);
+        const prDescriptionUntrimmed = prDescriptionRaw.replace(/\r+\n/g, '\n');
```
rousskov, line 626: I think this replacement should be moved to _parseRawLines() now. There is no reason to expose that specific step to _parse().

rousskov (approved):
LGTM with one trivial change.

src/MergeContext.js

PR #55, opened 2022-08-23
Fix invalid commit message detection

eduard-bagdasaryan:
Before this fix, Anubis incorrectly labeled unmergable PRs having valid
descriptions with M-failed-description. We should initialize
commitMessage field even for unmergeable/conflicted PRs, allowing to
perform required message checks and notify the user if the message is
invalid.


rousskov (approved):
Thank you for fixing this.

PR #56, opened 2022-08-29
Allow long header and trailer lines

eduard-bagdasaryan:
Limited the length of canonical (i.e. post-parsing/formatting)
'name: value' fields to 512 characters.

### DIFF
--- a/src/MergeContext.js
+++ b/src/MergeContext.js
@@ -494,6 +494,12 @@ class FieldsTokenizer
                         el.value.toUpperCase() === value.toUpperCase())) {
                 throw new Error(`duplicates are not allowed: ${line}`);
             }
+
+            const maxLength = 512;
+            const length = (name + ': ' + value).length;
+            if (length > maxLength)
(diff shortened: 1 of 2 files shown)


src/MergeContext.js
```diff
@@ -454,6 +454,11 @@ class BranchPosition
     }
 }
 
+function checkLineLength(line, limit) {
```
rousskov, line 457: If possible, I would use the default value here instead of duplicating it in callers:

```suggestion
function checkLineLength(line, limit = 72) {
```

Otherwise, we would need to move `72` from callers into a named DefaultLineLengthMax constant (to use in callers).

rousskov (approved):
LGTM with one minor suggested change. I also polished README a little; please check.

src/MergeContext.js
eduard-bagdasaryan: >  use the default value

Done.

eduard-bagdasaryan:
> polished README

It looks ok.

PR #57, opened 2022-10-13
Fix updateReference() returning 404 Not Found

eduard-bagdasaryan:
These failures started happening with/after the addition of GitHub
Actions. Our theory is that createCommit() returns SHA which is
unavailable for some time to subsequent API calls (due to those
unfinished actions). Now we wait in a do-while loop (up to ~30
seconds) until the SHA becomes available.

### DIFF
--- a/src/GitHubUtil.js
+++ b/src/GitHubUtil.js
@@ -127,8 +127,8 @@ async function getLabels(prNum) {
 // If requested and needed, retries until GitHub calculates PR mergeable flag.
 // Those retries, if any, are limited to a few minutes.
 async function getPR(prNum, awaitMergeable) {
-    const max = 64 * 1000 + 1; // ~2 min. overall
-    for (let d = 1000; d < max; d *= 2) {
+    const max = 64 * 1000; // limits total wait to ~2 minutes
+    for (let d = 1000; d <= max; d *= 2) {
         const pr = await getRawPR(prNum);
         // pr.mergeable is useless (and not calculated?) for a closed PR
         if (pr.mergeable !== null || pr.state === 'closed' || !awaitMergeable)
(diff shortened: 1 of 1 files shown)


src/GitHubUtil.js
```diff
@@ -322,6 +322,23 @@ async function updateReference(ref, sha, force) {
     return (await rateLimitedPromise(promise)).object.sha;
 }
 
+async function updateReference(ref, sha, force) {
+    const max = 16 * 1000 + 1; // ~30 sec. overall
+    for (let d = 1000; d < max; d *= 2) {
+        try {
+            await updateReferenceRaw(ref, sha, force);
```
rousskov, line 329: Missing successful termination handling?

```suggestion
            return await updateReferenceRaw(ref, sha, force);
```

src/GitHubUtil.js
```diff
@@ -322,6 +322,23 @@ async function updateReference(ref, sha, force) {
     return (await rateLimitedPromise(promise)).object.sha;
 }
 
+async function updateReference(ref, sha, force) {
+    const max = 16 * 1000 + 1; // ~30 sec. overall
+    for (let d = 1000; d < max; d *= 2) {
```
rousskov, line 327: Let's simplify and clarify, something along these lines:

```suggestion
    const max = 16 * 1000; // limits total wait to ~30 seconds
    for (let d = 1000; d <= max; d *= 2) {
```

rousskov (approved):
LGTM with changes.

rousskov:
Closing without merging: It turns out Anubis access token needs "workflow" permission to be able to update branches that contain GitHub Actions workflows (or something like that). I could not find any documentation about it (except one indirect [mention](https://github.blog/changelog/2020-06-03-personal-access-tokens-workflow-permission/)), but others suffering from the same "updateReference() returns 404" problem have received the corresponding advice from GitHub support and shared that solution with us. It appears to be working.

PR #60, opened 2023-03-08
Make some checks in PullRequest::_checkApproval() more strict/reliable

eduard-bagdasaryan:
Also removed debugging hook accidentally left behind by commit 63e257a.

### DIFF
--- a/src/MergeContext.js
+++ b/src/MergeContext.js
@@ -864,6 +864,8 @@ class PullRequest {
             usersVoted.push({reviewer: review.user.login, date: review.submitted_at, state: reviewState});
         }
 
+        // The loop could not just Approval.Block() on the first 'changes_requested'
+        // vote because that vote could have been dismissed later in the loop.
         const userRequested = usersVoted.find(el => el.state === 'changes_requested');
         if (userRequested !== undefined) {
             this._log("changes requested by " + userRequested.reviewer);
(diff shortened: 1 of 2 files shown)


eduard-bagdasaryan:
> we should probably refactor to completely remove usersVoted (by moving the above Suspend() call into the filtering loop).

I doubt we can get rid of usersVoted because of the reviews history that should be tracked. We cannot immediately suspend in the loop upon encountering 'changes_requested' state because it may be stale.

src/MergeContext.js
```diff
@@ -869,8 +869,10 @@ class PullRequest {
             this._log("changes requested by " + userRequested.reviewer);
```
rousskov, line 869: > Eduard at https://github.com/measurement-factory/anubis/pull/60#issuecomment-1460978061: I doubt we can get rid of usersVoted because of the reviews history that should be tracked. We cannot immediately suspend in the loop upon encountering 'changes_requested' state because it may be stale.

Good point, agreed! I missed the usersVoted.filter() call inside the loop. Please add the following comment above the current `userRequested` declaration to clarify why we are doing this here:

```C++
// The loop could not just Approval.Block() on the first 'changes_requested'
// vote because that vote could have been dismissed later in the loop.
```

rousskov (approved):
LGTM with one extra comment.

PR #61, opened 2023-05-11
Wait if staging branch status is being updated

eduard-bagdasaryan:
Generally we expect that if a GitHub API call returns success,
the operation has been completed on GitHub. However, this proved
to be wrong at least once for updateReference() call.

It is still an open question whether this problem relates to just
updateReverence() or all modifying operations (with PATCH/POST method).
We would need to come up with a more general approach in the second
case.

### DIFF
--- a/src/MergeContext.js
+++ b/src/MergeContext.js
@@ -451,6 +451,18 @@ class BranchPosition
         this._status = await GH.compareCommits(this._baseRef, this._featureRef);
     }
 
+    async waitUntil(aStatus) {
+        const max = 16 * 1000 + 1; // ~30 sec. overall
+        for (let d = 1000; d < max; d *= 2) {
+            await this.compute();
+            if (this._status === aStatus)
+                break;
+            Log.Logger.info("GitHub may be still updating branch status. Will retry in " + (d/1000) + " seconds");
+            await Util.sleep(d);
+        }
+        assert(this._status === aStatus);
+    }
+
     // feature > base:
     // the feature branch contains the base branch and some additional commits
     ahead() {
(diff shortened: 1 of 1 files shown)


rousskov (approved):
If you are OK with my adjustments, please feel free to merge and deploy.

src/MergeContext.js
```diff
@@ -449,6 +449,22 @@ class BranchPosition
...
+    }
+
+    async computeUntilAhead() {
+        const desiredStatus = "ahead";
+        // TODO: Stop (poorly) duplicating these Util.sleep() loops.
+        const longestSleep = 16 * 1000; // ms; ~30 seconds overall
+        let nextSleep = 0;
+        while (await this.compute() !== desiredStatus) {
+            if (nextSleep >= longestSleep)
+                throw new Error(`failed to reach the desired branch state; wanted ${desiredStatus} but got ${this._status} despite waiting for ${nextSleep/1000} seconds`);
```
eduard-bagdasaryan, line 462: Probably the overall spent time would be more informative - not just the last step duration. Basically, it should be lastStep*2-1. Do you think we should adjust it?

rousskov: > Probably the overall spent time would be more informative - not just the last step duration. 

I agree.

> Basically, it should be lastStep*2-1. 

Hm.. Is not it `(2^n-1)*a`, where n is the number of sleep calls, and a is the first sleep duration? For example, if we slept three times (for a, 2a, and 4a seconds), then the total for this n=3 case is 7a rather than `8a-1`.

> Do you think we should adjust it?

Yes, please, provided there is a simple way to report the actual total delay rather than the sum of Sleep() parameter values. AFAICT, that sum may not, in some cases, represent the actual time we delayed progress, even approximately, because our sleep might be interrupted and/or because compute() itself may take time.

eduard-bagdasaryan: Done at abfd49d.

rousskov: LGTM. Please feel free to merge and deploy.

PR #62, opened 2023-06-05
Ignore irrelevant check-runs

eduard-bagdasaryan:
Check Runs API returns the whole set of completed checks for the given
commit. In most cases the set does not contain duplicates because these
checks start only once for the commit. However, sometimes the same
commit may be tested multiple times (e.g., when the PR branch is
force-pushed with the same PR head commit). In this case, all results
are returned, making Anubis assert when checking tests uniqueness. To
avoid this, we need to select only the latest check runs from GitHub.

### DIFF
--- a/src/MergeContext.js
+++ b/src/MergeContext.js
@@ -973,8 +973,18 @@ class PullRequest {
                 statusChecks.addOptionalStatus(new StatusCheck(st));
         }
 
+        // Returns whole check runs history for the commit.
+        // We need only the freshest checks.
         const checkRuns = await GH.getCheckRuns(this._prHeadSha());
-        for (let st of checkRuns) {
+        // check.id is greater in newer checks
+        const sortedCheckRuns = checkRuns.sort((e1, e2) => parseInt(e1.id) - parseInt(e2.id));
+        let uniqueCheckRuns = [];
+        for (let i = sortedCheckRuns.length -1; i >= 0; --i) {
+            const check = sortedCheckRuns[i];
+            if (!uniqueCheckRuns.some(e => e.name === check.name))
+                uniqueCheckRuns.push(check);
+        }
+        for (let st of uniqueCheckRuns) {
             if (this._contextsRequiredByGitHubConfig.some(el => el.trim() === st.name.trim()))
                 statusChecks.addRequiredStatus(StatusCheck.FromCheckRun(st));
             else


src/MergeContext.js
```diff
@@ -973,8 +973,18 @@ class PullRequest {
                 statusChecks.addOptionalStatus(new StatusCheck(st));
         }
 
+        // Returns whole check runs history for the commit.
+        // We need only the freshest checks.
         const checkRuns = await GH.getCheckRuns(this._prHeadSha());
```
rousskov, line 978: Let's move the second part of this new comment down, to the code it describes.

```suggestion
        const checkRuns = await GH.getCheckRuns(this._prHeadSha());
        // Filter out stale/older checks, assuming that check.id is greater in newer checks.
```

src/MergeContext.js
```diff
@@ -973,8 +973,18 @@ class PullRequest {
...
-        for (let st of checkRuns) {
+        // check.id is greater in newer checks
```
rousskov, line 979: I merged this line into the comment above it.
```suggestion
```

src/MergeContext.js
```diff
@@ -973,8 +973,18 @@ class PullRequest {
...
+        const sortedCheckRuns = checkRuns.sort((e1, e2) => parseInt(e1.id) - parseInt(e2.id));
+        let uniqueCheckRuns = [];
+        for (let i = sortedCheckRuns.length -1; i >= 0; --i) {
```
rousskov, line 982: I hope this explicit loop will go away with other changes, but if it does not, please add a space:

```suggestion
        for (let i = sortedCheckRuns.length - 1; i >= 0; --i) {
```

src/MergeContext.js
```diff
@@ -973,8 +973,18 @@ class PullRequest {
...
+            const check = sortedCheckRuns[i];
```
rousskov, line 983: Feels odd to sort the array one way and then work extra hard to iterate it in the reverse order. Please see whether you can adjust sortedCheckRuns so that freshest checks come first and then use forEach() to iterate that array. Something along these untested lines may work:

```suggestion
        const sortedCheckRuns = checkRuns.sort((e1, e2) => parseInt(e2.id) - parseInt(e1.id));
        let uniqueCheckRuns = [];
        sortedCheckRuns.forEach(check => {
```

rousskov (commented):
LGTM with some minor polishing touches.

src/MergeContext.js

eduard-bagdasaryan (commented):
Done (b338ea2).

PR #63, opened 2023-07-24
Rate limiting for 'paginated' results

eduard-bagdasaryan:
Addressed a TODO from 5024fd2.

### DIFF
--- a/src/GitHubUtil.js
+++ b/src/GitHubUtil.js
@@ -1,9 +1,14 @@
 const assert = require('assert');
 const Config = require('./Config.js');
-const GitHub = require('@octokit/rest')({
-    timeout: Config.requestTimeout(),
-    host: 'api.github.com',
-    version: '3.0.0'
+const { Octokit }  = require("@octokit/rest");
+const GitHub = new Octokit({
+    auth: Config.githubToken(),
+    request: {
+        agent: undefined,
+        fetch: undefined,
+        timeout: Config.requestTimeout(),
+    },
+    baseUrl: 'https://api.github.com'
 });
 const Util = require('./Util.js');
 const Log = require('./Logger.js');
@@ -11,8 +16,6 @@ const Log = require('./Logger.js');
 const ErrorContext = Util.ErrorContext;
 const commonParams = Util.commonParams;
 const logApiResult = Log.logApiResult;
-
-const GitHubAuthentication = { type: 'token', username: Config.githubUserLogin(), token: Config.githubToken() };
 
 function defaultAppender(allPages, aPage) {
     allPages.data = allPages.data.concat(aPage.data);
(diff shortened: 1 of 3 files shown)


src/GitHubUtil.js
```diff
@@ -163,9 +172,10 @@ async function getCommits(branch, since) {
     params.sha = branch; // sha or branch to start listing commits from
     params.since = since;
 
-    const commits = await GitHub.paginate(GitHub.rest.repos.listCommits, params);
-    logApiResult(getCommits.name, params, {commits: commits.length});
-    return commits;
+    const pages = await GitHub.paginate(GitHub.rest.repos.listCommits, params, (response) => response);
+    const result = combinePages(pages);
```
rousskov, line 176: These two lines are repeated many times and contain some low-level logic/magic. If possible, please move them to an internal paginatedGet() function. If you do move them, it is possible that you will not need/want the combinePages() helper anymore (and will inline its code).

src/GitHubUtil.js
```diff
@@ -101,19 +109,20 @@ async function getReviews(prNum) {
...
     params.pull_number = prNum;
 
-    const reviews = await GitHub.paginate(GitHub.rest.pulls.listReviews, params);
-    logApiResult(getReviews.name, params, {reviews: reviews.length});
-    // TODO: implement rateLimitedPromise() for 'paginated' results
-    return reviews;
+    const pages = await GitHub.paginate(GitHub.rest.pulls.listReviews, params, (response) => response);
+    const result = combinePages(pages);
+    logApiResult(getReviews.name, params, {reviews: result.data.length});
+    return await rateLimitedPromise(result);
```
rousskov, line 115: This does not look like the right rate limiting place to me: I would expect us to pause after each page rather than at the end of all pages. Am I missing something?

src/GitHubUtil.js
```diff
@@ -101,19 +109,20 @@ async function getReviews(prNum) {
     let params = commonParams();
     params.pull_number = prNum;
 
-    const reviews = await GitHub.paginate(GitHub.rest.pulls.listReviews, params);
-    logApiResult(getReviews.name, params, {reviews: reviews.length});
-    // TODO: implement rateLimitedPromise() for 'paginated' results
-    return reviews;
+    const pages = await GitHub.paginate(GitHub.rest.pulls.listReviews, params, (response) => response);
```
rousskov, line 112: Do we need that no-op(?) lambda at the end? I see some paginate() [examples](https://github.com/octokit/octokit.js#pagination) that do not use it, but I am not sure I may be comparing apples to oranges. If it is needed, let's add a source code comment to document why.

rousskov (changes requested):

eduard-bagdasaryan (commented):
Done (095519c).

rousskov (approved):
LGTM! Please merge and deploy when you are comfortable with the new code.

PR #65, opened 2024-02-06
Ignore unchanged PRs

eduard-bagdasaryan:
Often, many open PRs are not modified for a long time. Do not waste
resources analyzing them over and over. This optimization is not applied
to PRs delayed by Anubis itself.

### DIFF
--- a/src/PrMerger.js
+++ b/src/PrMerger.js
@@ -6,6 +6,18 @@ const GH = require('./GitHubUtil.js');
 const Util = require('./Util.js');
 const MergeContext = require('./MergeContext.js');
 
+class PrScanResult {
+    constructor() {
+        this.scanDate = new Date(); // the scan starting time
+        this.delay = null; // the next scan should start in this number of milliseconds
+        this.unchangedPrs = []; // PRs that may be ignored in the next scan
+    }
+
+    isStillUnchanged(pr) {
+        return this.unchangedPrs.some(el => el.number === pr.number && el.updated_at === pr.updated_at);
+    }
+}
+
 // A single Anubis processing step:
 // Updates and, to the extent possible, advances each open PR. Once.
 class PrMerger {
@@ -14,21 +26,21 @@ class PrMerger {
         this._total = 0; // the number of open PRs received from GitHub
         this._errors = 0; // the number of PRs with processing failures
         this._ignored = 0; // the number of ignored PRs (due to human markings)
+        this._ignoredUnchanged = 0; // the number of ignored PRs (due to lack of PR modifications)
         this._todo = null; // raw PRs to be processed
     }
 
     // Implements a single Anubis processing step.
     // Returns suggested wait time until the next step (in milliseconds).
-    async execute() {
+    async execute(oldScanResult) {
         Logger.info("runStep running");
+        let scanResult = new PrScanResult();
 
         this._todo = await GH.getOpenPrs();
         this._total = this._todo.length;
         Logger.info(`Received ${this._total} PRs from GitHub:`, this._prNumbers());
 
         await this._determineProcessingOrder(await this._current());
-
-        let minDelay = null;
 
         let somePrWasStaged = false;
         while (this._todo.length) {
@@ -41,11 +53,32 @@ class PrMerger {
                     continue;
                 }
 
+                const clearedForMerge = rawPr.labels.some(el => el.name === Config.clearedForMergeLabel());
+
+                // There are usually few 'cleared' PRs (i.e., ready for merge), so do not ignore them.
+                // This should help handle sutiations where some of such 'cleared' PRs were updated just after
+                // this scan has been started. Other (non-cleared) PRs (that are not going to be merged now anyway)
+                // can be ignored until the next scan.
+                // TODO: also handle a situation when the PR becomes 'cleared' just after this scan began.
+                if (!clearedForMerge && oldScanResult && oldScanResult.isStillUnchanged(rawPr)) {
+                    Logger.info(`Ignoring PR${rawPr.number} because it has not changed since the previous scan of ${oldScanResult.scanDate.toISOString()}`);
+                    this._ignoredUnchanged++;
+                    scanResult.unchangedPrs.push(rawPr);
+                    continue;
+                }
+
                 const result = await MergeContext.Process(rawPr, somePrWasStaged);
                 assert(!somePrWasStaged || !result.prStaged());
                 somePrWasStaged = somePrWasStaged || result.prStaged();
-                if (result.delayed() && (minDelay === null || minDelay > result.delayMs()))
-                    minDelay = result.delayMs();
+                if (result.delayed() && (scanResult.delay === null || scanResult.delay > result.delayMs()))
+                    scanResult.delay = result.delayMs();
+
+                // Delayed PRs will be handled by the bot itself in the future, do not ignore them.
+                if (!result.delayed()) {
+                    const unmodifiedDurationMs = new Date() - new Date(rawPr.updated_at);
+                    if (unmodifiedDurationMs >= Config.prUnchangedDurationMin())
+                        scanResult.unchangedPrs.push(rawPr);
+                }
             } catch (e) {
                 Log.LogError(e, "PrMerger.runStep");
                 this._errors++;
@@ -55,8 +88,8 @@ class PrMerger {
         if (this._errors)
             throw new Error(`Failed to process ${this._errors} out of ${this._total} PRs.`);
 
-        Logger.info("Successfully handled all " + this._total + " PRs, processed/ignored: " + (this._total - this._ignored) + "/" + this._ignored);
-        return minDelay;
+        Logger.info("Successfully handled all " + this._total + " PRs, processed/ignored: " + (this._total - this._ignored) + "/" + this._ignored + "+" + this._ignoredUnchanged);
+        return scanResult;
     }
 
     // a string enumerating PR numbers of _todo PRs
@@ -122,10 +155,13 @@ class PrMerger {
 } // PrMerger
 
 // promises to process all PRs once, hiding PrMerger from callers
-function Step() {
+function Step(oldScanResult) {
     let mergerer = new PrMerger();
-    return mergerer.execute();
+    return mergerer.execute(oldScanResult);
 }
 
-module.exports = Step;
+module.exports = {
+    Step: Step,
+    PrScanResult: PrScanResult
+};
 
--- a/src/PrMerger.js
+++ b/src/PrMerger.js
@@ -9,14 +9,20 @@ const MergeContext = require('./MergeContext.js');
 class PrScanResult {
     constructor() {
         this.scanDate = new Date(); // the scan starting time
-        this.delay = null; // the next scan should start in this number of milliseconds
-        this.unchangedPrs = []; // PRs that may be ignored in the next scan
+        this.awakePrs = []; // PRs that were not delayed during the scan
     }
 
-    isStillUnchanged(pr) {
-        return this.unchangedPrs.some(el => el.number === pr.number && el.updated_at === pr.updated_at);
+    isStillUnchanged(rawPr) {
+        const pr = this.awakePrs.find(el => el.number === rawPr.number && el.updated_at === rawPr.updated_at);
+        if (pr === undefined)
+            return false;
+        const unmodifiedDurationMs = new Date() - new Date(pr.updated_at);
+        return unmodifiedDurationMs > 1000*3600;
     }
 }
+
+// PrScanResult produced by the latest PrMerger.execute() call
+let _LastScan = null;
 
 // A single Anubis processing step:
 // Updates and, to the extent possible, advances each open PR. Once.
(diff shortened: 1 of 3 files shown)


src/PrMerger.js
```diff
@@ -41,11 +53,32 @@ class PrMerger {
...
+                    continue;
+                }
+
                 const result = await MergeContext.Process(rawPr, somePrWasStaged);
                 assert(!somePrWasStaged || !result.prStaged());
                 somePrWasStaged = somePrWasStaged || result.prStaged();
-                if (result.delayed() && (minDelay === null || minDelay > result.delayMs()))
-                    minDelay = result.delayMs();
+                if (result.delayed() && (scanResult.delay === null || scanResult.delay > result.delayMs()))
+                    scanResult.delay = result.delayMs();
```
rousskov, line 74: I do not understand why we need to remember minDelay (beyond this function call lifetime). I would expect this code to stay intact and scanResult to have no delay field.

src/PrMerger.js
```diff
@@ -6,6 +6,18 @@ const GH = require('./GitHubUtil.js');
 const Util = require('./Util.js');
 const MergeContext = require('./MergeContext.js');
 
+class PrScanResult {
+    constructor() {
+        this.scanDate = new Date(); // the scan starting time
+        this.delay = null; // the next scan should start in this number of milliseconds
+        this.unchangedPrs = []; // PRs that may be ignored in the next scan
```
rousskov, line 13: To simplify and clarify (user) logic, please store all non-delayed PRs in this array:

```suggestion
        this.awakePrs = []; // PRs that were not delayed during the scan
```

src/PrMerger.js
```diff
@@ -6,6 +6,18 @@ const GH = require('./GitHubUtil.js');
...
+    }
+
+    isStillUnchanged(pr) {
+        return this.unchangedPrs.some(el => el.number === pr.number && el.updated_at === pr.updated_at);
```
rousskov, line 17: Please refactor to apply one hour delta from the original issue specs: PRs that were updated delta milliseconds ago should be considered "potentially changed" even if their updated_at field has not changed. This is done to avoid ignoring same-timestamp updates. BTW, if there is a monotonically increased field that we can use instead of updated_at here, we should use that field and avoid this delta.

AFAICT, you can hard-code that delta here, in this method; no need for prUnchangedDurationMin().

src/PrMerger.js
```diff
@@ -14,22 +26,22 @@ class PrMerger {
         this._total = 0; // the number of open PRs received from GitHub
         this._errors = 0; // the number of PRs with processing failures
         this._ignored = 0; // the number of ignored PRs (due to human markings)
+        this._ignoredUnchanged = 0; // the number of ignored PRs (due to lack of PR modifications)
```
rousskov, line 29:
```suggestion

        // here, "ignored" or "skipped" means "not given to MergeContext.Process()"
        this._ignored = 0; // the number of PRs skipped due to human markings; TODO: Rename to _ignoredAsMarked
        this._ignoredAsUnchanged = 0; // the number of PRs skipped due to lack of PR updates

```

src/PrMerger.js
```diff
@@ -41,11 +53,32 @@ class PrMerger {
...
+
+                // There are usually few 'cleared' PRs (i.e., ready for merge), so do not ignore them.
+                // This should help handle sutiations where some of such 'cleared' PRs were updated just after
+                // this scan has been started. Other (non-cleared) PRs (that are not going to be merged now anyway)
+                // can be ignored until the next scan.
+                // TODO: also handle a situation when the PR becomes 'cleared' just after this scan began.
+                if (!clearedForMerge && oldScanResult && oldScanResult.isStillUnchanged(rawPr)) {
+                    Logger.info(`Ignoring PR${rawPr.number} because it has not changed since the previous scan of ${oldScanResult.scanDate.toISOString()}`);
+                    this._ignoredUnchanged++;
+                    scanResult.unchangedPrs.push(rawPr);
```
rousskov, line 66: Instead of pushing here and below while skipping this push for ignoredByMergeBotsLabel() PRs above, please push _all_ PRs, before checking ignoredByMergeBotsLabel(). And then, after the Process() call below, remove (i.e. filter out) delayed PRs.

src/PrMerger.js
```diff
@@ -41,11 +53,32 @@ class PrMerger {
...
+
                 const result = await MergeContext.Process(rawPr, somePrWasStaged);
                 assert(!somePrWasStaged || !result.prStaged());
                 somePrWasStaged = somePrWasStaged || result.prStaged();
-                if (result.delayed() && (minDelay === null || minDelay > result.delayMs()))
-                    minDelay = result.delayMs();
+                if (result.delayed() && (scanResult.delay === null || scanResult.delay > result.delayMs()))
+                    scanResult.delay = result.delayMs();
+
+                // Delayed PRs will be handled by the bot itself in the future, do not ignore them.
```
rousskov, line 76:
```suggestion
                // delayed PRs will be processed (when the delay expires) even if not updated
```

rousskov (changes requested):

eduard-bagdasaryan (commented):
Done (aefbcac).

src/PrMerger.js
```diff
@@ -123,9 +171,13 @@ class PrMerger {
...
+    try {
+        let mergerer = new PrMerger();
+        return mergerer.execute();
+    } catch (e) {
+        _LastScan = null;
+        throw e;
+    }
 }
 
 module.exports = Step;
```
rousskov, line 183:
```suggestion
module.exports = Step;

```

src/PrMerger.js
```diff
@@ -35,17 +57,37 @@ class PrMerger {
...
                 const result = await MergeContext.Process(rawPr, somePrWasStaged);
                 assert(!somePrWasStaged || !result.prStaged());
                 somePrWasStaged = somePrWasStaged || result.prStaged();
                 if (result.delayed() && (minDelay === null || minDelay > result.delayMs()))
                     minDelay = result.delayMs();
+
+                // delayed PRs will be processed (when the delay expires) even if not updated
+                if (result.delayed())
+                    scanResult.awakePrs = scanResult.awakePrs.filter(el => el.number !== rawPr.number);
+
```
rousskov, line 90:
```suggestion
                    currentScan.forgetPr(rawPr);
```

And please assert, inside the new method, that the given PR _is_ found.

src/PrMerger.js
```diff
@@ -6,21 +6,43 @@ const GH = require('./GitHubUtil.js');
...
+    isStillUnchanged(rawPr) {
+        const pr = this.awakePrs.find(el => el.number === rawPr.number && el.updated_at === rawPr.updated_at);
+        if (pr === undefined)
+            return false;
+        const unmodifiedDurationMs = new Date() - new Date(pr.updated_at);
+        return unmodifiedDurationMs > 1000*3600;
+    }
+}
+
+// PrScanResult produced by the latest PrMerger.execute() call
```
rousskov, line 24:
```suggestion
// PrScanResult produced by the last successfully finished PrMerger.execute() call
```

src/PrMerger.js
```diff
@@ -123,9 +171,13 @@ class PrMerger {
...
 function Step() {
-    let mergerer = new PrMerger();
-    return mergerer.execute();
+    try {
+        let mergerer = new PrMerger();
+        return mergerer.execute();
+    } catch (e) {
+        _LastScan = null;
+        throw e;
+    }
```
rousskov, line 180: This is awkward. I suspect forgetPr() will become forgetDelayedPr() so that it can manage the delay, and we will eventually end up with something like this:

```suggestion
    const lastScan = _LastScan;
    _LastScan = null;
    const mergerer = new PrMerger(lastScan);
    _LastScan = mergerer.execute();
    return _LastScan.minDelay();
```

... but we can wait for the next iteration to decide.

rousskov (changes requested):
Good progress, thank you. I think we are approaching the final solution here, but we still need one or two non-trivial iterations.

src/PrMerger.js
```diff
@@ -6,21 +6,43 @@ const GH = require('./GitHubUtil.js');
...
+    constructor() {
+        this.scanDate = new Date(); // the scan starting time
+        this.awakePrs = []; // PRs that were not delayed during the scan
+    }
+
+    isStillUnchanged(rawPr) {
+        const pr = this.awakePrs.find(el => el.number === rawPr.number && el.updated_at === rawPr.updated_at);
+        if (pr === undefined)
+            return false;
+        const unmodifiedDurationMs = new Date() - new Date(pr.updated_at);
```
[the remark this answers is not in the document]
eduard-bagdasaryan: > the longer a single scan runs, the fewer PRs will be considered unmodified. 

No, I would say the more a single scan runs, the more PRs wil be considered unmodified (e.g., if the scan runs more than an hour then at least one PR will be definitely unmodified). But I agree that we should pass the scan start time and treat all PRs equally in this respect.

src/PrMerger.js
eduard-bagdasaryan: Ok.

src/PrMerger.js
eduard-bagdasaryan: Ok, let's postpone until the next iteration.

eduard-bagdasaryan (commented):
Done (235c5da).

src/PrMerger.js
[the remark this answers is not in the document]
rousskov: > > the longer a single scan runs, the fewer PRs will be considered unmodified.

> I would say the more a single scan runs, the more PRs wil be considered unmodified.

Yes, I got it backwards, probably due to the "negative" function name. Sorry.

src/PrMerger.js
```diff
@@ -35,17 +57,37 @@ class PrMerger {
...
                 }
 
+                const clearedForMerge = rawPr.labels.some(el => el.name === Config.clearedForMergeLabel());
+
+                // There are usually few 'cleared' PRs (i.e., ready for merge), so do not ignore them.
+                // This should help handle sutiations where some of such 'cleared' PRs were updated just after
+                // this scan has been started. Other (non-cleared) PRs (that are not going to be merged now anyway)
+                // can be ignored until the next scan.
+                // TODO: also handle a situation when the PR becomes 'cleared' just after this scan began.
+                if (!clearedForMerge && _LastScan && _LastScan.isStillUnchanged(rawPr)) {
```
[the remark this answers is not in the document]
rousskov: > > would not the updated_at field of the just-cleared PR already reflect the update?!
> 
> What if the PR is updated after `GH.getOpenPrs()` but before the loop iteration on this PR?

... then we would not know that the PR was cleared -- getOpenPrs() would not contain the clearance label.

From GitHub and relevant Anubis code point of view, the clearance label is similar to any other label (and labeling itself is similar to other PR updates that keep a PR open!). I would like us to be able to rely on bare isStillUnchanged() check because if we cannot, then this code will not work correctly in cases other than cleared for merged PRs.

AFAICT, this clearedForMerge check will _speed up_ cleared PR processing in the following cases, but they ought to be very rare (or, if GitHub properly implements atomic PR metadata updates, nonexistent!). The red flags raised by this rare-case optimization are just not worth it IMO.

1. GitHub changes PR label.
2. GitHub sends half-updated PR metadata to Anubis (in response to a getOpenPrs() request).
3. GitHub changes PR updated_at (as the result of the label change in item 1).
4. GitHub sends a notification about the fully-updated PR.

src/PrMerger.js
```diff
@@ -6,14 +6,44 @@ const GH = require('./GitHubUtil.js');
...
+            return false; // PR has changed since this scan
+        // treat recently updated PRs as changed PRs thus factoring in a
+        // (slight) possibility of same-timestamp changes
+        const unmodifiedDurationMs = freshScanDate - new Date(savedRawPr.updated_at);
+        return unmodifiedDurationMs > 1000*3600;
+    }
+
+    forgetPr(rawPr) {
+        assert(this.awakePrs.find(el => el.number === rawPr.number));
+        this.awakePrs = this.awakePrs.filter(el => el.number !== rawPr.number);
```
rousskov, line 29: Just to speed it up a little:

```suggestion
        const oldPrs = this.awakePrs;
        this.awakePrs = this.awakePrs.filter(el => el.number !== rawPr.number);
        assert(oldPrs.length === this.awakePrs.length + 1); // one PR was removed
```

src/PrMerger.js
rousskov: I think now is the right time to do this, moving minDelay maintenance into PrMerger.

rousskov (changes requested):

src/PrMerger.js
eduard-bagdasaryan: Done.

eduard-bagdasaryan (commented):
Done (bf9a882).

src/PrMerger.js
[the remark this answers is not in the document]
eduard-bagdasaryan: Removed the 'clearedForMerge' condition (bf9a882).

rousskov (approved):
LGTM, thank you.

PR #67, opened 2024-03-04
Add workflow_run and check_run GitHub event handlers

eduard-bagdasaryan:
We should have added these event handlers together with check-runs API
support (63e257a). PRs did not stuck probably because the bot was
tickled by other/older checks that usually finish later (e.g., Jenkins).


rousskov (approved):
Thank you for fixing this!

PR #68, opened 2024-05-26
Automatically count required staging checks

eduard-bagdasaryan:
Now Anubis requests the list of required staging checks from GitHub
'stating branch' settings instead of maintaining its own
config::staging_checks parameter. That parameter was problematic because
it was usually out of sync with the actutal list of required checks.
With new approach, the list of required staging checks should still be
kept up-to-date. However, it became easier because it must include CI
checks only (GitHub check runs are maintained automatically).

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ are satisfied:
   branch is out of date with its target branch.
 * The PR is not a
   [draft](https://docs.github.com/en/github/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-stage-of-a-pull-request).
-* All the _required_ checks have succeeded on the PR branch:
+* All the _required_ PR checks have succeeded on the PR branch:
   * If _all_ checks have succeeded, then GitHub says "All checks have
     passed" next to a green check mark:
     ![](./docs/images/all_passed.png)
@@ -47,7 +47,7 @@ are satisfied:
     ![](./docs/images/required_passed.png)
   
     Unfortunately, GitHub also displays the above message when some
-    required checks have failed. To determine whether all the required
+    required PR checks have failed. To determine whether all the required PR
     checks have succeeded when some checks have failed, expand the check
     list using the "show all checks" link.
 * The PR is approved for merging (see below for voting rules).
@@ -69,8 +69,8 @@ algorithm:
    section further below.
 2. Reset the staging (a.k.a. "auto") branch to the PR staging commit.
    The previous state of the staging branch is ignored.
-3. Wait for GitHub to report exactly `config::staging_checks` CI test
-   results for the staging branch. The project CI infrastructure is
+3. Wait for GitHub to report that all the _required_ staging checks have
+   succeeded on the staging branch. The project CI infrastructure is
    expected to auto-test the staging branch on every change. A check
    failure is a PR-specific step failure -- in GitHub terminology, all
    staging checks are currently deemed "required". If discovered, any
@@ -126,7 +126,7 @@ request state:
   for merging, successfully completed the initial merging steps that are
   supposed to trigger CI tests, and is now waiting for the CI test
   results. The bot removes this label when it notices that all
-  `config::staging_checks` tests have completed.
+  required staging checks have succeeded on the staging branch.
 * `M-passed-staging-checks`: Similar to the GitHub "green check" mark
   for the staging branch commit (but ignores failures of optional
   checks). Usually only visible when the bot is running in staging-only
@@ -384,7 +384,6 @@ All configuration fields are required.
 *core_developers* | The comma-separated list of login=id pairs, specifying GitHub login and ID for core developers. | ""
 *voting_delay_min*| The minimum merging age of a PR. Younger PRs are not merged, regardless of the number of votes. The PR age string should comply with [timestring](https://github.com/mike182uk/timestring) parser. | "2d"
 *voting_delay_max* | The maximum merging age of a PR that has fewer than `config::sufficient_approvals` votes. The PR age string should comply with [timestring](https://github.com/mike182uk/timestring) parser. | "10d"
-*staging_checks*| The expected number of CI tests executed against the staging branch. | 2
 *approval_url*| The URL associated with an approval status test description. | ""
 *logger_params* | A JSON-formatted parameter list for the [Bunyan](https://github.com/trentm/node-bunyan) logging library [constructor](https://github.com/trentm/node-bunyan#constructor-api). | <pre>{<br>    "name": "anubis",<br>    "streams": [ ... ]<br>}</pre>
 
--- a/src/Config.js
+++ b/src/Config.js
@@ -29,7 +29,6 @@ class ConfigOptions {
         assert(this._sufficientApprovals > 1);
         this._votingDelayMax = timestring(conf.voting_delay_max, 'ms');
         this._votingDelayMin = timestring(conf.voting_delay_min, 'ms');
-        this._stagingChecks = conf.staging_checks;
         this._loggerParams = conf.logger_params;
         this._approvalUrl = conf.approval_url;
         this._coreDevelopers = conf.core_developers;
@@ -96,7 +95,6 @@ class ConfigOptions {
     sufficientApprovals() { return this._sufficientApprovals; }
     votingDelayMax() { return this._votingDelayMax; }
     votingDelayMin() { return this._votingDelayMin; }
-    stagingChecks() { return this._stagingChecks; }
     loggerParams() { return this._loggerParams; }
     // returns a Map of (login,id) pairs
     coreDeveloperIds() { return this._coreDeveloperIds; }
--- a/src/MergeContext.js
+++ b/src/MergeContext.js
@@ -174,8 +174,7 @@ class StatusChecks
     // final() should count all statuses.
 
     // expectedStatusCount:
-    //   for staged commits: the bot-configured number of required checks (Config.stagingChecks()),
-    //   for PR commits: GitHub configured number of required checks (requested from GitHub)
+    // GitHub configured number of required checks (requested from GitHub)
     // context: either "PR" or "Staging";
     // contextsRequiredByGitHubConfig: the GitHub protected branch checks in a format of a list of string contexts
     constructor(expectedStatusCount, context, contextsRequiredByGitHubConfig) {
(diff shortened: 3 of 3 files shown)


src/MergeContext.js
```diff
@@ -234,7 +229,7 @@ class StatusChecks
 
     // no more required status changes or additions are expected
     final() {
-        return (this.requiredStatuses.length >= this.expectedStatusCount) &&
+        return (this.requiredStatuses.length >= this._contextsRequiredByGitHubConfig.length) &&
```
eduard-bagdasaryan, line 232: this.requiredStatuses.length should be <= this._contextsRequiredByGitHubConfig.length because this.requiredStatuses only gets statuses that are among this._contextsRequiredByGitHubConfig. So we can probably assert the '<=' condition here and use '==' instead for the returning condition.

README.md
```diff
@@ -69,13 +69,9 @@ algorithm:
    section further below.
 2. Reset the staging (a.k.a. "auto") branch to the PR staging commit.
    The previous state of the staging branch is ignored.
-3. Wait for GitHub to report exactly `config::staging_checks` CI test
-   results for the staging branch. The project CI infrastructure is
-   expected to auto-test the staging branch on every change. A check
-   failure is a PR-specific step failure -- in GitHub terminology, all
-   staging checks are currently deemed "required". If discovered, any
-   extra check is considered a configuration problem (not related to any
```
eduard-bagdasaryan, line 77: Now all 'extra' checks will be treated as optional (i.e., the checks that are not 'required').

src/MergeContext.js
```diff
@@ -1016,10 +1018,13 @@ class PullRequest {
             statusChecks.addOptionalStatus(new StatusCheck(st));
         }
```
eduard-bagdasaryan, line 1019: _getPrStatuses() and _getStagingStatuses() look almost the same now, except for this dealing with 'copied' statuses (although the diff does not show it clearly).

src/MergeContext.js
```diff
@@ -997,16 +1002,13 @@ class PullRequest {
     async _getStagingStatuses() {
         const combinedStagingStatuses = await GH.getStatuses(this._stagedSha());
         const genuineStatuses = combinedStagingStatuses.statuses.filter(st => !st.description.endsWith(Config.copiedDescriptionSuffix()));
-        assert(genuineStatuses.length <= Config.stagingChecks());
-        let statusChecks = new StatusChecks(Config.stagingChecks(), "Staging", this._contextsRequiredByGitHubConfig);
-        // all genuine checks, except for PR approval, are 'required'
-        // PR approval may be either 'required' or 'optional' (depending on the GitHub settings)
+        let statusChecks = new StatusChecks("Staging", this._contextsRequiredByGitHubConfigStaging);
         for (let st of genuineStatuses) {
```
rousskov, line 1006: If you agree with it, please leave this code polishing comment for a followup PR to address:

```suggestion
        // TODO: Move this loop inside StatusChecks constructor by adding
        // a third constructor parameter to pass genuineStatuses (here) and
        // combinedPrStatus.statuses (in similar _getPrStatuses() code).
        for (let st of genuineStatuses) {
```

If you agree with the "move this loop inside StatusChecks constructor" part but see problems with "by ..." part than just add the first part:

```suggestion
        // TODO: Move this loop inside StatusChecks constructor to reuse
        // it here and in similar _getPrStatuses() code.
        for (let st of genuineStatuses) {
```

src/MergeContext.js
rousskov: Yes, I noticed that too and suggested a TODO in another change request...

src/MergeContext.js
```diff
@@ -1016,10 +1018,13 @@ class PullRequest {
             statusChecks.addOptionalStatus(new StatusCheck(st));
```
rousskov, line 1018: Do we still need to add these copied statuses to `statusChecks`? I may be missing something, but I am hoping that this copying was required when we were just _counting_ statuses. Now that we can explicitly check that every status required by humans (via GitHub auto branch configuration) is successful, perhaps we no longer need to do this?

Alternatively, if we still have to do this, perhaps the earlier loop with its `_contextsRequiredByGitHubConfigStaging` condition would take care of it if we stop filtering by copiedDescriptionSuffix() just before that loop?

Per commit 8510a56 description, we have to copy PR statuses to staged commit to "overcome GitHub protection for staged commits". My understanding is that we still need to copy to staged commit, for that old reason. In other words, _supplyStagingWithPrRequired() is still needed. This change request is about this statusChecks.addOptionalStatus() call and not _supplyStagingWithPrRequired(). I assume the two are independent.

src/MergeContext.js
```diff
@@ -997,16 +1002,13 @@ class PullRequest {
...
-        let statusChecks = new StatusChecks(Config.stagingChecks(), "Staging", this._contextsRequiredByGitHubConfig);
-        // all genuine checks, except for PR approval, are 'required'
-        // PR approval may be either 'required' or 'optional' (depending on the GitHub settings)
+        let statusChecks = new StatusChecks("Staging", this._contextsRequiredByGitHubConfigStaging);
         for (let st of genuineStatuses) {
             const check = new StatusCheck(st);
-            if (check.context.trim() === Config.approvalContext())
-                statusChecks.setApprovalStatus(check);
-            else
+            if (this._contextsRequiredByGitHubConfigStaging.some(el => el.trim() === st.context.trim()))
```
rousskov, line 1008: AFAICT, this change of `if` statement condition implies that GitHub configuration should now require PR Approval status check for the auto branch if (and only if) GitHub configuration requires PR Approval status check for the targeted base branch. In other words, the two branch configurations (targeted/base and auto/staging) should match as far as classifying PR Approval status check as "required" or "optional".

If that is indeed a requirement, we should assert somewhere that GitHub configuration complies with it!

src/MergeContext.js
```diff
@@ -1652,7 +1657,8 @@ class PullRequest {
 
         await this._loadStaged();
         await this._loadRawPr(); // requires this._loadStaged()
-        await this._loadRequiredContexts();
+        this._contextsRequiredByGitHubConfigBase = await this._loadRequiredContexts(this._prBaseBranch());
+        this._contextsRequiredByGitHubConfigStaging = await this._loadRequiredContexts(Config.stagingBranch());
```
rousskov, line 1661: Please rename PR's _loadRequiredContexts() to _loadRequiredContextsFor() and then move these two assignments into  _loadRequiredContexts() function (that will do nothing else but set these two fields). Doing so will effectively undo PR change this comment is attached to, restoring this code original "style", and restoring synchronization with the "requires..." comment below.

```suggestion
        await this._loadRequiredContexts();
```

rousskov (changes requested):
Thank you for this improvement! I hope we can go a step further, but I am not sure. Please see change requests for details.

src/MergeContext.js
rousskov: BTW, the following Squid configuration means (AFAICT) that Squid runs PR checks for staged commits. Perhaps we can simplify Anubis code if we let projects decide what to test and what to require instead of assuming that a project does not run PR checks for staged commits but does require them? In other words, can we adopt the following simple principle: If a project wants to require a check, it must run that check (instead of relying on Anubis to copy that check from one branch/commit to another).

```yaml
name: PR commit

on:
  push:
    # test commits on this branch and staged commits
    branches: [ "master", "auto" ]
```

src/MergeContext.js
rousskov: This change request may become obsolete if "we do not need this check copying anymore" [change request](https://github.com/measurement-factory/anubis/pull/68#discussion_r2107826018) results in PR Approval check not copied anymore. The text below assumes that we still want to copy for some reason.

----

AFAICT:

* Before this PR, setApprovalStatus() set/added PR Approval check as required iff GitHub **PR branch** was configured to require that check (via this._approvalIsRequired).
* After this PR, this code sets/adds PR Approval check as required iff GitHub **auto branch** was configured to require that check.

I did not check all the details, and it is possible that the above difference effectively disappears somewhere later, but, on the surface, it looks like a potentially important change.

> What happens if they do not match?

If we require them to match (for some valid reason that we need to establish), then we assert that they do match sometime before merging a PR.

src/MergeContext.js
rousskov: > IIRC, we added this hack due to a Jenkins quirk: PR had one integral Jenkins check (added to the master protection), while staged commit got multiple Jenkins checks (added by another Jenkins plugin) - so we had to add that integral check before merge to satisfy the master protection requirement.

Let's assume that we have removed copying logic today, and then Jenkins or similar setup was resurrected tomorrow. Would we still be able to accommodate that Jenkins setup by instructing the Project to
* configure PR/master branch protection to require "integral check" _and_
* configure staging/auto branch protection to _not_ require that check?

I hope the answer to my question is something like "yes, we can still support Jenkins-like setups after this PR because Anubis enforces that all required PR checks (including 'integral check') are still green before merging a PR that passed all required staging checks (that may not contain an 'integral check')". Is that how this PR would work if we remove "that copying logic"?

eduard-bagdasaryan: > if we remove "that copying logic"

As we had discussed, we cannot remove this copying logic. I adjusted it to override the existing auto-branch optional checks with PR required checks.

src/MergeContext.js
eduard-bagdasaryan: > Please rename PR's _loadRequiredContexts()

Done.

eduard-bagdasaryan (commented):
Done and merged from master, resolving conflicts.

src/Config.js
```diff
@@ -90,14 +89,14 @@ class ConfigOptions {
     baseUrl() { return 'https://api.github.com'; }
     stagingBranchPath() { return "heads/" + this._stagingBranch; }
     stagingBranch() { return this._stagingBranch; }
+    stagingRuleSetName() { return "StagingRuleSet"; }
```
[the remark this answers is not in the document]
rousskov: > we need to search by a known ruleset name (implemented) or a ID

AFAICT, ideally, we should request [ruleset details](https://docs.github.com/en/rest/repos/rules?apiVersion=2022-11-28#get-a-repository-ruleset) (using ruleset ID discovered during the previous "request the list of available RuleSets" step) and then use `conditions.ref_name` field in those details to see if it ends with stagingBranchPath().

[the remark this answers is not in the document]
eduard-bagdasaryan: Ok, done (987d61). Iterating over the RuleSet list (requesting details via a separate API call) may require more API calls though (in case there are many RuleSets configured). But for now (a single RuleSet configured) it should be the same.

PR #70, opened 2025-03-11
Do not ignore PRs with M-cleared-for-merge

eduard-bagdasaryan:
We should bypass the idle PR optimization in this case
because cleared-for-merge PRs may be ready for merge
without additional GitHub activity.

### DIFF
--- a/src/PrMerger.js
+++ b/src/PrMerger.js
@@ -14,6 +14,10 @@ class PrScanResult {
     }
 
     isStillUnchanged(freshRawPr, freshScanDate) {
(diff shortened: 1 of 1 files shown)


rousskov (approved):
I have adjusted PR title/description and polished isStillUnchanged() implementation a bit. If you are OK with these changes, please deploy ASAP.

I hope they will fix the current situation where, AFAICT, none of multiple cleared PRs can merge because one cleared PR gets ignored while its staging tests are running and a another cleared PR is staged, invalidating those running tests. Then that other cleared PR gets ignored, etc., etc.

PR #73, opened 2025-05-08
Update Octokit library to 21.1.0 version

eduard-bagdasaryan:
This fixes the Octokit bug:

     Unexpected end of JSON input #649

The bug occurred from time to time during getStatuses() calls.
Also ported the code from CommonJS to ESM package style,
since Octokit >21.0.0 requires ESM package management.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "A bot that merges GitHub pull requests",
   "main": "Main.js",
   "dependencies": {
-    "@octokit/rest": "^18.12.0",
+    "@octokit/rest": "^21.1.0",
     "bunyan": "^1.8.12",
     "github-webhook-handler": "^0.6.0",
     "parse-github-url": "^0.3.2",
@@ -22,5 +22,6 @@
     "url": "git+https://github.com/measurement-factory/anubis.git"
   },
   "author": "The Measurement Factory",
-  "license": "Apache-2.0"
+  "license": "Apache-2.0",
+  "type": "module"
 }
--- a/src/Config.js
+++ b/src/Config.js
@@ -1,6 +1,6 @@
-const fs = require('fs');
-const timestring = require('timestring');
-const assert = require('assert');
+import fs from 'fs';
+import timestring from 'timestring';
+import assert from 'assert';
 
 class ConfigOptions {
     constructor(fname) {
@@ -141,4 +141,4 @@ class ConfigOptions {
 const configFile = process.argv.length > 2 ? process.argv[2] : './config.json';
 const Config = new ConfigOptions(configFile);
 
-module.exports = Config;
+export default Config;
--- a/src/MergeContext.js
+++ b/src/MergeContext.js
@@ -1,9 +1,8 @@
-const assert = require('assert');
-const Config = require('./Config.js');
-const Log = require('./Logger.js');
-const GH = require('./GitHubUtil.js');
-const Util = require('./Util.js');
-
+import assert from 'assert';
+import Config from './Config.js';
+import * as Log from './Logger.js';
+import * as GH from './GitHubUtil.js';
+import * as Util from './Util.js';
 
 // Process() outcome
 class ProcessResult
@@ -1795,13 +1794,8 @@ class PullRequest {
 }
 
 // promises to update/advance the given PR, hiding PullRequest from callers
-function Process(rawPr, banStaging) {
+export default function Process(rawPr, banStaging) {
     let pr = new PullRequest(rawPr, banStaging);
     return pr.process();
 }
-
-
-module.exports = {
-    Process: Process
-};
 
--- a/src/PrMerger.js
+++ b/src/PrMerger.js
@@ -1,10 +1,10 @@
-const assert = require('assert');
-const Config = require('./Config.js');
-const Log = require('./Logger.js');
+import assert from 'assert';
+import Config from './Config.js';
+import * as Log from './Logger.js';
 const Logger = Log.Logger;
-const GH = require('./GitHubUtil.js');
-const Util = require('./Util.js');
-const MergeContext = require('./MergeContext.js');
+import * as GH from './GitHubUtil.js';
+import * as Util from './Util.js';
+import Process from './MergeContext.js';
 
 class PrScanResult {
     constructor(prs) {
@@ -135,7 +135,7 @@ class PrMerger {
                     continue;
                 }
 
-                const result = await MergeContext.Process(rawPr, somePrWasStaged);
+                const result = await Process(rawPr, somePrWasStaged);
                 assert(!somePrWasStaged || !result.prStaged());
                 somePrWasStaged = somePrWasStaged || result.prStaged();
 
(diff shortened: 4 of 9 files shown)


src/Config.js
```diff
@@ -1,6 +1,6 @@
-const fs = require('fs');
-const timestring = require('timestring');
-const assert = require('assert');
+import fs from 'fs';
+import timestring from 'timestring';
+import assert from 'assert';
```
rousskov, line 3:
```
eslint `git ls-files . | grep '[.]js$'`

/home/alex/Edit/anubis/src/Config.js
  1:1  error  Parsing error: 'import' and 'export' may appear only with 'sourceType: module'

/home/alex/Edit/anubis/src/GitHubUtil.js
  1:1  error  Parsing error: 'import' and 'export' may appear only with 'sourceType: module'

/home/alex/Edit/anubis/src/Logger.js
  1:1  error  Parsing error: 'import' and 'export' may appear only with 'sourceType: module'

/home/alex/Edit/anubis/src/Main.js
  1:1  error  Parsing error: 'import' and 'export' may appear only with 'sourceType: module'

/home/alex/Edit/anubis/src/MergeContext.js
  1:1  error  Parsing error: 'import' and 'export' may appear only with 'sourceType: module'

/home/alex/Edit/anubis/src/PrMerger.js
  1:1  error  Parsing error: 'import' and 'export' may appear only with 'sourceType: module'

/home/alex/Edit/anubis/src/RepoMerger.js
  1:1  error  Parsing error: 'import' and 'export' may appear only with 'sourceType: module'

/home/alex/Edit/anubis/src/Util.js
  1:1  error  Parsing error: 'import' and 'export' may appear only with 'sourceType: module'

✖ 8 problems (8 errors, 0 warnings)
```

Is there a way to still support eslint after switching to `import`? I know that Daft uses `import` declarations but does not trigger eslint parsing errors similar to above.

rousskov (commented):
I have updated PR description a little. Please check, especially the documented reason for switching from require() to `import` declarations.

src/Config.js
eduard-bagdasaryan: I could not make our current ancient eslint version (4.14.0) to deal with this, so I had to update it to the latest version (9.26.0). That required some other adjustments, e.g., new configuration style (they abandoned .eslintrc since 9.0.0), that conversion is made by a tool. To fix the "Parsing error: 'import' and 'export'" problems, I just set the sourceType parameter "module". Also I ran it against our sources and fixed several formatting problems (like spaces and semicolons). It complained a lot about the camel case, and  I had to configure `properties : "never"` so that it did not complain about object properties not conforming to the camel case (we need these properties to configure Octokit API calls).

rousskov: Ouch! I do support eslint upgrade in principle, but there are many changes here and some raise additional red flags and questions that we need to discuss/check. Let's step back and do this eslint upgrade properly. I see two primary paths forward:

1. Revert eslint-specific commit in this PR (i.e. 84b0000aa6abb35d122b53d7433ea07797c81cd6). Add an XXX/TODO about eslint to this PR description. Merge this PR without eslint fixes. Do a dedicated "upgrade eslint" PR afterwards.

2. Do a dedicated eslint upgrade PR first. Revert eslint-specific commit in this PR (i.e. 84b0000aa6abb35d122b53d7433ea07797c81cd6). Merge this PR without eslint fixes (because those fixes will already be in master by then).

Please peak the path you think works best overall. If you do not see a significant difference, let's go with option 1.

> that conversion is made by a tool

Regardless of the PR merge order, for the eslint-dedicated PR, please do not start with the equivalent of "all in one" commit 84b0000aa6abb35d122b53d7433ea07797c81cd6. Instead, commit the result of that automated or tool-driven conversion first. That way we can see what we are manually changing/customizing in subsequent commits (i.e. beyond the automated conversion). I presume that automated conversion is meant to preserve original settings to the extent possible.

eduard-bagdasaryan: Ok, I removed eslint changes from this PR.

src/MergeContext.js
```diff
@@ -1795,13 +1794,8 @@ class PullRequest {
 }
 
 // promises to update/advance the given PR, hiding PullRequest from callers
-function Process(rawPr, banStaging) {
+export default function Process(rawPr, banStaging) {
```
rousskov, line 1797: If possible without a lot of changes, let's force callers to specify explicitly what they want to import:

```suggestion
export function Process(rawPr, banStaging) {
```

src/PrMerger.js
```diff
@@ -278,7 +278,7 @@ class PrMerger {
 } // PrMerger
 
 // promises to process all PRs once, hiding PrMerger from callers
-async function Step(prIds) {
+export default async function Step(prIds) {
```
rousskov, line 281:
```suggestion
export async function Step(prIds) {
```

Please apply similar changes to other cases that can be converted without heroic efforts.

src/PrMerger.js
```diff
@@ -1,10 +1,10 @@
-const assert = require('assert');
-const Config = require('./Config.js');
-const Log = require('./Logger.js');
+import assert from 'assert';
+import Config from './Config.js';
+import * as Log from './Logger.js';
 const Logger = Log.Logger;
```
rousskov, line 4: Please use the following order of directives:

1. import Anubis modules, alphabetical whole-line order ("sort lines" in most editors should get this right)
2. empty line (if any Anubis modules were imported above)
3. import external modules, alphabetical whole-line order ("sort lines" in most editors should get this right)
4. empty line (if any external modules were imported above)

FWIW, Daft tries to use the same approach (but there are violations because we used a different approach earlier and because none of this is currently enforced).

package.json
```diff
@@ -22,5 +22,6 @@
     "url": "git+https://github.com/measurement-factory/anubis.git"
   },
   "author": "The Measurement Factory",
-  "license": "Apache-2.0"
+  "license": "Apache-2.0",
+  "type": "module"
```
rousskov, line 26: Please move this new "type" property to position it between "description" and "main" entries above.

src/PrMerger.js
```diff
@@ -1,10 +1,10 @@
...
-const GH = require('./GitHubUtil.js');
-const Util = require('./Util.js');
-const MergeContext = require('./MergeContext.js');
+import * as GH from './GitHubUtil.js';
+import * as Util from './Util.js';
+import Process from './MergeContext.js';
```
rousskov, line 7: Let's preserve (more informative) MergeContext.Process() call further below:

```suggestion
import * as MergeContext from './MergeContext.js';
```

rousskov (changes requested):
Just some minor polishing touches...

rousskov:
I also adjusted PR title/description. Please check.

src/PrMerger.js
eduard-bagdasaryan: Done.

src/MergeContext.js

src/PrMerger.js
eduard-bagdasaryan: Ok.

src/PrMerger.js
eduard-bagdasaryan: Ok. I left only Config as default - to avoid Config.Config everywhere...

eduard-bagdasaryan (commented):
Done.

PR #74, opened 2025-06-02
Upgrade eslint to v9.27.0

eduard-bagdasaryan:
Modern ESLint uses a different configuration style. The corresponding
configuration file was generated using migrate-config tool.

We now configure ESLint to ignore camelCase violations in object
properties because there are many such violations in imported Octokit
library (e.g., `pull_number` and `issue_number`), and ESLint cannot be
configured to only ignore properties of imported objects.

Also fixed (benign) eslint-detected code problems.


rousskov (approved):
LGTM, assuming that camelCase change is the only configuration change we know about. If you are OK with my PR title/description changes, please feel free to merge and deploy. Thank you.

PR #78, opened 2026-03-30
Fix PR association with workflow_run/check_run events

eduard-bagdasaryan:
GitHub has been removing PR information from check_run and worflow_run
events (especially for forks), returning empty pull_requests array in
the events payload. Anubis could not get this information to associate
the events with monitoring PRs:

    "check_run event: no PR for ..."
    "workflow_run event: no PR for ..."

Now Anubis extracts PR branch name from these events, and searches
an open PR with the same branch name. This approach has been already
applied for 'status' and 'push' events.

### DIFF
--- a/src/Main.js
+++ b/src/Main.js
@@ -107,9 +107,10 @@ function handleCheckEvent(name, e) {
     if (e.head_branch === Config.stagingBranch()) {
         return Util.PrId.Sha(e.head_sha);
     }
-    // check/workflow events for other non-PR branches (e.g., master) are not associated with PRs
-    Logger.info(`${name} event: no PR for`, e.head_sha);
-    return [];
+
+    // Check/workflow events for other branches.
+    // Will check for PR association later.
+    return Util.PrId.BranchList([e.head_branch], null);
 }
 
 // https://docs.github.com/ru/webhooks/webhook-events-and-payloads#workflow_run
--- a/src/Main.js
+++ b/src/Main.js
@@ -104,12 +104,15 @@ function handleCheckEvent(name, e) {
         return [];
     }
 
+    if (e.head_branch === undefined) {
+        Logger.info(`${name} event: no e.head_branch`, e.head_sha);
+        return [];
+    }
+
     if (e.head_branch === Config.stagingBranch()) {
         return Util.PrId.Sha(e.head_sha);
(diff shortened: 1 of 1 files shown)


src/Main.js
```diff
@@ -107,9 +107,10 @@ function handleCheckEvent(name, e) {
     if (e.head_branch === Config.stagingBranch()) {
         return Util.PrId.Sha(e.head_sha);
     }
-    // check/workflow events for other non-PR branches (e.g., master) are not associated with PRs
-    Logger.info(`${name} event: no PR for`, e.head_sha);
-    return [];
+
+    // Check/workflow events for other branches.
+    // Will check for PR association later.
```
rousskov: Let's remove these comments. For "other branches", the code already speaks for itself. A for the "Will check" promise, it looks odd that we only add that promise here even though the `stagingBranch()` code above does not have a (visible) PR association (yet) either.

```suggestion
```

src/Main.js
```diff
@@ -107,9 +107,10 @@ function handleCheckEvent(name, e) {
...
+    return Util.PrId.BranchList([e.head_branch], null);
```
rousskov, line 116: Please check that `e.head_branch` is defined. I know we have not done that before, but we have not "fully" used an undefined value before either -- we only excluded it as "other", probably with a console warning about using an undefined value in equality comparison.

Please place that check before the first `e.head_branch` use above and return `[]` after logging a `"no e.head_branch"` debugging note.

rousskov, line 116: > PR description: Now Anubis extracts PR branch name from these events, and searches
an open PR with the same branch name. This approach has been already applied for 'status' and 'push' events.

Nice!

I am concerned that `check_run-completed.json` event dump does not have a `check_run.head_branch` field. It has `check_run.check_suite.head_branch` instead. Does this event handler get `check_run.check_suite` part of that event?

rousskov (changes requested):

src/Main.js
eduard-bagdasaryan: Done.

src/Main.js
eduard-bagdasaryan: removed.

rousskov (approved):

PR #79, opened 2026-05-25
Wait for status checks to be copied before attempting to merge

eduard-bagdasaryan:
After we apply required status checks to a staged commit and
subsequently attempt to merge that commit, GitHub may return an error
that the staged commit lacks those required checks. Evidently, GitHub
responds to a createStatus() call before that call has an effect, and it
may take a few seconds for the checks to be actually applied. Now we
wait up to 2 minutes and report an error if the status is still missing.

Also removed ErrorContext class and fixed the related exception
handlers. ErrorContext was used in old Anubis versions in promisificated
wrappers of Octokit API calls. After updating Octokit to v21 at c1fac18,
ErrorContext was eliminated from all wrappers but one getPR() call.
Now the last use case is gone and hence, the class itself too.

Also fixed broken exception handlers with ErrorContext checks
(unreachable since c1fac18). We should analyze these 404 and 422 errors
using Octokit-generated RequestError instead. The new error handling
differs from the old one: Previously, we treated 404 and 422 errors, as
well as all non-HTTP errors, as 'notFound()' or 'unprocessale()',
depending on the context.  Now we consider only genuine 404 and 422
errors as such.  However, this behavior change should probably not
matter now since the old behavior has been broken since c1fac18.

### DIFF
--- a/src/GitHubUtil.js
+++ b/src/GitHubUtil.js
@@ -270,26 +270,6 @@ export async function getComments(prNum) {
     return comments;
 }
 
-// XXX: remove if not needed, since the "required_status_checks" api call sometimes
-// does not work(?) for organization repositories (returns 404 Not Found).
-//async function getProtectedBranchRequiredStatusChecks(branch) {
-//    let params = commonParams();
-//    params.branch = branch;
-//    const promise = new Promise( (resolve, reject) => {
-//      GitHub.authenticate(GitHubAuthentication);
-//      GitHub.repos.getProtectedBranchRequiredStatusChecks(params, (err, res) => {
-//          if (err) {
-//             reject(new ErrorContext(err, getProtectedBranchRequiredStatusChecks.name, params));
-//             return;
-//          }
-//          const result = {checks: res.data.contexts.length};
-//          logApiResult(getProtectedBranchRequiredStatusChecks.name, params, result);
-//          resolve(res);
-//      });
-//    return (await rateLimitedPromise(promise)).contexts;
-//    });
-//}
-
 export async function createStatus(sha, state, targetUrl, description, context) {
     assert(!Config.dryRun());
 
--- a/src/GitHubUtil.js
+++ b/src/GitHubUtil.js
@@ -81,21 +81,38 @@ export async function getLabels(prNum) {
     return await rateLimitedPromise(result);
 }
 
+/// For waitFor() code to signal that more iterations are needed
+const TryAgain = Symbol("TryAgain");
+
+// (Re)runs `code` while it returns a TryAgain value, waiting between retries.
+// Meant for `code` that makes GitHub API call(s) that may not succeed on the
+// first try.
+async function waitFor(description, code) {
+    // this limits total wait time to about 2 minutes
+    const singleWaitMax = 64; // seconds
+    for (let singleWait = 1; singleWait <= singleWaitMax; singleWait *= 2) {
+        const result = await code();
+        if (result !== TryAgain) {
+            Log.Logger.info(`Done waiting for ${description}`);
+            return result;
+        }
+        Log.Logger.info(`Still waiting for ${description}. Will retry in ${singleWait} seconds`);
+        await Util.sleep(singleWait*1000);
+    }
+    throw new ErrorContext(`Timed out waiting for ${description}`, waitFor.name);
+}
+
 // Gets PR metadata from GitHub
 // If requested and needed, retries until GitHub calculates PR mergeable flag.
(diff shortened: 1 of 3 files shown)


src/GitHubUtil.js
```diff
@@ -274,6 +274,26 @@ export async function getComments(prNum) {
 //    });
 //}
 
+async function getStatus(ref, context) {
```
rousskov: Please avoid code duplication using the following sketch as a starting point:

```js
/// For waitFor() code to signal that more iterations are needed
const TryAgain = Symbol("TryAgain");

// (Re)runs `code` while it returns a TryAgain value, waiting between retries.
// Meant for `code` that makes GitHub API call(s) that may not succeed on the
// first try.
async function waitFor(description, code) {
    // this limits total wait time to about 2 minutes
    const singleWaitMax = 64; // seconds
    for (let singleWait = 1; singleWait <= singleWaitMax; singleWait *= 2) {
        const result = await code();
        if (result !== TryAgain) {
            Log.Logger.info(`Done waiting for ${description}`);
            return result;
        }
        Log.Logger.info(`Still waiting for ${description}. Will retry in ${singleWait} seconds`);
        await Util.sleep(singleWait*1000);
    }
    throw new ErrorContext(`Timed out waiting for ${description}`, waitFor.name);
}
```

Here is how that new function may be used to implement the two old functions that need it:

```js
export function getPR(prNum, awaitMergeable) {
    return waitFor('GitHub to calculate mergeable attribute', async () => {
        const pr = await getRawPR(prNum);
        await rateLimitedPromise(pr).context;
        // pr.mergeable is useless (and not calculated?) for a closed PR
        if (pr.mergeable !== null || pr.state === 'closed' || !awaitMergeable)
            return pr;
        return TryAgain;
    }
}
```

```js
export async function createStatus(sha, state, targetUrl, description, context) {
    assert(!Config.dryRun());

    let params = commonParams();
    params.sha = sha;
    params.state = state;
    params.target_url = targetUrl;
    params.description = description;
    params.context = context;

    const result = await GitHub.rest.repos.createCommitStatus(params);
    logApiResult(createStatus.name, params, {context: result.data.context});
    await rateLimitedPromise(result).context;

    const checkParams = commonParams();
    checkParams.ref = sha;
    return waitFor(`GitHub to create ${context} status for the ${sha} commit`, async () => {
        const checkResult = await GitHub.rest.repos.getCombinedStatusForRef(checkParams);
        await rateLimitedPromise(checkResult).context;
        const statuses = checkResult.data.statuses;
        if (statuses.some(st => st.context === context)) {
            return statuses;
        }
        return TryAgain;
    }
}
```

rousskov (changes requested):

src/GitHubUtil.js
```diff
@@ -274,6 +274,26 @@ export async function getComments(prNum) {
...
+        const statuses = result.data.statuses;
+        if (statuses.some(st => st.context === context)) {
+            Log.Logger.info("Verified that GitHub has successfully created " + context + " status for the " + ref + " commit.");
+            return true;
+        }
+        Log.Logger.info("GitHub is still applying " + context + " status to the " + ref + " commit. Will retry in " + (d/1000) + " seconds");
+        await Util.sleep(d);
+    }
+
+    return Promise.reject(new ErrorContext("Timed out waiting for GitHub to create commit status",
```
eduard-bagdasaryan: As far as we are going to abandon this obsolete 'reject' approach, I would suggest to remove our ErrorContext class and use standard Error instead. We used ErrorContext in each API wrapper in old Anubis versions (e.g., as in commented getProtectedBranchRequiredStatusChecks()) and some high-level callers also handled specially this exception type (e.g., MergeContext._removeFromGitHub()). After moving to a new Octokit version and adjusting our API wrappers, ErrorContext is gone in all wrappers but getPR(), rendering also the special higher-level handling unreachable. I think we need to fix these cases, analyzing [RequestError](https://github.com/octokit/octokit.js/#request-error-handling) from Octokit library there.

src/GitHubUtil.js
eduard-bagdasaryan: Adjusted (37d075c7).

src/GitHubUtil.js
eduard-bagdasaryan: Done (1b6fe7e).

rousskov (approved):
I have updated PR description a bit. Please add an "Also removed the ErrorContext class..." paragraph describing the effects of those side changes, as we discussed.

PR #81, opened 2026-06-29
Create PR merge commit instead of relying on GitHub

eduard-bagdasaryan:
We used to rely on GitHub's ability to generate PR merge commits any
time PR's head or base branch changes.  Currently GitHub does not keep
merge commits in sync with the base branch (see recent commit c2ded1e).
There are several ways to force GitHub to re-generate a merge commit:

* switch PR branch to point to a random branch and then switch it back;
* merge PR base branch into PR branch;
* open and close the PR.

We rejected all these options because they are unreliable (e.g., failing
to switch PR base branch back leaves the PR in a bad state) and/or
create noise on PR page, strange notifications and overall bad UX.

Instead of relying on GitHub doing that, Anubis now creates a PR merge
commit, using a dedicated branch with a configurable name
(`bot_merge_branch`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -379,6 +379,7 @@ All configuration fields are required.
 *staged_run*| A boolean option to enable staging-only mode where the bot performs all the merging steps up to (but not including) the target branch update. Eligible PRs are merged into and tested on the staging branch but are never merged into their target branches. Staging-only mode prevents any target branch modifications by the bot. TODO: Check that the PR target branch is not the configured staging branch, setting `M-failed-other` if needed. | false
 *guarded_run*| Enables staging-only mode (see `config::staged_run`) for PRs without a `M-cleared-for-merge` label. Has no effect on PRs with that label. While `config::staged_run` blocks target branch modifications, this option allows them for a human-designated subset of PRs. | false
 *staging_branch* | The name of the bot-maintained git branch used for testing PR changes as if they were merged into their target branch. | auto
+*merge_branch* | The name of the bot-maintained git branch used for keeping the result of merging the PR branch into the base branch. | auto__
 *necessary_approvals* | The minimal number of core developers required for a PR to be merged. PRs with fewer votes are not merged, regardless of their age. | 1
 *sufficient_approvals* | The minimal number of core developers required for a PR to be merged fast (i.e., without waiting for `config::voting_delay_max`) | 2
 *core_developers* | The comma-separated list of login=id pairs, specifying GitHub login and ID for core developers. | ""
--- a/README.md
+++ b/README.md
@@ -248,9 +248,7 @@ formatting but strips any trailing whitespace.
 Anubis assembles the staging git commit (and, hence, in case of a successful
 PR merge, the target branch commit) from the following parts:
 
-* `tree`: The git tree object of the PR [merge
-  commit](https://developer.github.com/v3/pulls/#get-a-single-pull-request)
-  created by GitHub.
+* `tree`: The git tree object from the PR merge commit.
 * `parents`: The HEAD commit of the PR base branch.
 * `message`: See the "Commit message" subsection for details.
 * `author`: The name and email from the author object of the PR merge commit
@@ -379,7 +377,7 @@ All configuration fields are required.
 *staged_run*| A boolean option to enable staging-only mode where the bot performs all the merging steps up to (but not including) the target branch update. Eligible PRs are merged into and tested on the staging branch but are never merged into their target branches. Staging-only mode prevents any target branch modifications by the bot. TODO: Check that the PR target branch is not the configured staging branch, setting `M-failed-other` if needed. | false
 *guarded_run*| Enables staging-only mode (see `config::staged_run`) for PRs without a `M-cleared-for-merge` label. Has no effect on PRs with that label. While `config::staged_run` blocks target branch modifications, this option allows them for a human-designated subset of PRs. | false
 *staging_branch* | The name of the bot-maintained git branch used for testing PR changes as if they were merged into their target branch. | auto
-*merge_branch* | The name of the bot-maintained git branch used for keeping the result of merging the PR branch into the target branch. | auto__
+*bot_merge_branch* | The name of the bot-maintained git branch that is very similar to GitHub-maintained PR merge branch. GitHub no longer updates its PR merge branches (i.e. refs/pull/PULL_REQUEST_NUMBER/merge references) in some important use cases, so Anubis uses its own branch (with a configurable name) instead. The last commit on Anubis bot_merge_branch (PR merge commit) is the result of merging PR branch into PR target branch, as a regular git merge command would do. Anubis uses the code tree from that last commit when creating a squashed commit on the staging_branch. The primary difference between the two branches is that staging_branch ends in a squashed (i.e. single-parent) commit rather than a regular two-parent merge commit; after both branches are updated, a git diff staging_branch..bot_merge_branch command produces no output because both branches contain the same code tree. | auto__
 *necessary_approvals* | The minimal number of core developers required for a PR to be merged. PRs with fewer votes are not merged, regardless of their age. | 1
 *sufficient_approvals* | The minimal number of core developers required for a PR to be merged fast (i.e., without waiting for `config::voting_delay_max`) | 2
 *core_developers* | The comma-separated list of login=id pairs, specifying GitHub login and ID for core developers. | ""
--- a/README.md
+++ b/README.md
@@ -379,7 +379,7 @@ All configuration fields are required.
 *staged_run*| A boolean option to enable staging-only mode where the bot performs all the merging steps up to (but not including) the target branch update. Eligible PRs are merged into and tested on the staging branch but are never merged into their target branches. Staging-only mode prevents any target branch modifications by the bot. TODO: Check that the PR target branch is not the configured staging branch, setting `M-failed-other` if needed. | false
 *guarded_run*| Enables staging-only mode (see `config::staged_run`) for PRs without a `M-cleared-for-merge` label. Has no effect on PRs with that label. While `config::staged_run` blocks target branch modifications, this option allows them for a human-designated subset of PRs. | false
 *staging_branch* | The name of the bot-maintained git branch used for testing PR changes as if they were merged into their target branch. | auto
-*merge_branch* | The name of the bot-maintained git branch used for keeping the result of merging the PR branch into the base branch. | auto__
+*merge_branch* | The name of the bot-maintained git branch used for keeping the result of merging the PR branch into the target branch. | auto__
 *necessary_approvals* | The minimal number of core developers required for a PR to be merged. PRs with fewer votes are not merged, regardless of their age. | 1
 *sufficient_approvals* | The minimal number of core developers required for a PR to be merged fast (i.e., without waiting for `config::voting_delay_max`) | 2
 *core_developers* | The comma-separated list of login=id pairs, specifying GitHub login and ID for core developers. | ""
--- a/README.md
+++ b/README.md
@@ -135,7 +135,7 @@ request state:
   staging results are no longer fresh/applicable.
 * `M-abandoned-staging-checks`: Anubis discovered a stale _staging commit_.
   This usually happens when either the PR state changes (e.g., the author
-  commits new code) or the staging_branch is unexpectedly modified. After
+  commits new code) or the `staging_branch` is unexpectedly modified. After
   labeling the PR, the bot ignores any failed or unfinished tests associated
   with that stale commit. This label was added so that the developer observing
   the PR page on GitHub knows why Anubis ignored (failed, unfinished, or
@@ -248,13 +248,13 @@ formatting but strips any trailing whitespace.
 Anubis assembles the staging git commit (and, hence, in case of a successful
 PR merge, the target branch commit) from the following parts:
 
-* `tree`: The git tree object from the PR merge commit.
+* `tree`: The git tree object from the bot PR merge commit.
 * `parents`: The HEAD commit of the PR base branch.
 * `message`: See the "Commit message" subsection for details.
-* `author`: The name and email from the author object of the PR merge commit
+* `author`: The name and email from the author object of the GitHub PR merge commit
   unless overwritten by the `Authored-by` header field value from the PR
   description header.
-* `author.date`: The date from the author object of the PR merge commit.
+* `author.date`: The date from the author object of the GitHub PR merge commit.
 
 Note that the individual commits on the PR branch (including their messages
 and authors) are ignored. The bot is effectively performing a squash merge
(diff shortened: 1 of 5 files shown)


src/MergeContext.js
```diff
@@ -1332,6 +1329,8 @@ class PullRequest {
 
     _prBaseBranchPath() { return "heads/" + this._prBaseBranch(); }
 
+    _prHeadBranch() { return this._rawPr.head.ref; }
```
rousskov: Please move to where `_prHeadSha()` is defined.

src/MergeContext.js
```diff
@@ -1496,20 +1495,11 @@ class PullRequest {
...
-        if (this._prMergeable()) {
-            const mergeSha = await GH.getReference(this._mergePath());
-            this._mergeCommit = await GH.getCommit(mergeSha);
-            defaultAuthor = this._mergeCommit.author;
-            stageable = true;
-        } else {
-            const headCommit = await GH.getCommit(this._prHeadSha());
-            defaultAuthor = headCommit.author;
-        }
+        const headCommit = await GH.getCommit(this._prHeadSha());
```
rousskov: In what typical cases this change results in a staged commit author change? Do we need to change Anubis documentation accordingly?

src/MergeContext.js
```diff
@@ -1496,20 +1495,11 @@ class PullRequest {
...
+        let defaultAuthor = headCommit.author;
```
```suggestion
        const defaultAuthor = headCommit.author;
```

src/MergeContext.js
```diff
@@ -1537,18 +1527,6 @@ class PullRequest {
...
-        const treeShaIsFresh = this._stagedCommit.tree.sha === this._mergeCommit.tree.sha;
-        this._log("staged commit tree sha freshness: " + treeShaIsFresh);
-        if (!treeShaIsFresh)
-            return false;
-
-        assert(this._stagedCommit.parents.length === 1);
-        const stagedCommitParentSha = this._stagedCommit.parents[0].sha;
-        const parentIsFresh = this._mergeCommit.parents.some(p => p.sha === stagedCommitParentSha);
-        this._log("staged commit parent freshness: " + parentIsFresh);
-        if (!parentIsFresh)
```
rousskov, line 1549: Can we check that staged commit parents are still fresh?

rousskov (changes requested):

src/MergeContext.js
eduard-bagdasaryan: I undid this change.

src/MergeContext.js

src/MergeContext.js
eduard-bagdasaryan: I restored both checks: they now check against our custom mergeCommit.

eduard-bagdasaryan (commented):
Done (e5a1585).

src/GitHubUtil.js
```diff
@@ -217,6 +217,20 @@ export async function updateReference(ref, sha, force) {
     return await rateLimitedPromise(result);
 }
 
+export async function mergeBranch(base, head) {
```
rousskov: It is difficult to guess parameter order correctly when looking at calling code. Let's reorder parameters to allow for a more explicit name:

```suggestion
export async function mergeAintoB(head, base) {
```

... and adjust the caller accordingly.

src/GitHubUtil.js
```diff
@@ -217,6 +217,20 @@ export async function updateReference(ref, sha, force) {
...
+    assert(!Config.dryRun());
+    let params = commonParams();
+    params.base = base;
+    params.head = head;
+    params.commit_message = `Merged ${head} into ${base}`;
```
rousskov: Let's be more unique/explicit here in case somebody stumbles upon these "internal" Anubis commits in the official repository:

```suggestion
    params.commit_message = `Merged ${head} into ${base} to create a staged commit code tree`;
```

src/MergeContext.js
```diff
@@ -1499,9 +1499,11 @@ class PullRequest {
...
         let stageable = false;
         if (this._prMergeable()) {
-            const mergeSha = await GH.getReference(this._mergePath());
-            this._mergeCommit = await GH.getCommit(mergeSha);
-            defaultAuthor = this._mergeCommit.author;
+            const defaultMergeSha = await GH.getReference(this._defaultMergePath());
+            const defaultMergeCommit = await GH.getCommit(defaultMergeSha);
+            // Github regenerates its ephemeral merge commit each time the PR branch is updated.
+            // Author attributes will be up to date with the latest PR branch changes.
+            defaultAuthor = defaultMergeCommit.author;
```
```suggestion
            const oldGhMergeBranchSha = await GH.getReference(this._oldGhMergeBranchPath());
            const oldGhMergeBranchCommit = await GH.getCommit(oldGhMergeBranchSha);
            // GitHub does reset its merge branch each time the PR branch is updated,
            // keeping author attributes up to date with the latest PR branch changes.
            defaultAuthor = oldGhMergeBranchCommit.author;
```

README.md
```diff
@@ -379,6 +379,7 @@ All configuration fields are required.
 *staged_run*| A boolean option to enable staging-only mode where the bot performs all the merging steps up to (but not including) the target branch update. Eligible PRs are merged into and tested on the staging branch but are never merged into their target branches. Staging-only mode prevents any target branch modifications by the bot. TODO: Check that the PR target branch is not the configured staging branch, setting `M-failed-other` if needed. | false
 *guarded_run*| Enables staging-only mode (see `config::staged_run`) for PRs without a `M-cleared-for-merge` label. Has no effect on PRs with that label. While `config::staged_run` blocks target branch modifications, this option allows them for a human-designated subset of PRs. | false
 *staging_branch* | The name of the bot-maintained git branch used for testing PR changes as if they were merged into their target branch. | auto
+*merge_branch* | The name of the bot-maintained git branch used for keeping the result of merging the PR branch into the base branch. | auto__
```
rousskov: `staging_branch` description and a lot of other README.md places use "target branch" terminology. There is only one other "base branch" usage AFAICT. Let's continue to use "target" in this PR (and then possibly followup with another dedicated PR to replace "target" with "base" if "base" is a better term).

Let's rename `merge_branch` configuration option to `bot_merge_branch` to highlight the difference from GitHub-maintained merge branches.

I would also like to change the proposed description of `bot_merge_branch` to the following or similar text:

The name of the bot-maintained git branch that is very similar to GitHub-maintained [PR merge
  branch](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-pull-requests#pull-request-refs-and-merge-branches). GitHub no longer updates its PR merge branches (i.e. `refs/pull/PULL_REQUEST_NUMBER/merge` references) in some important use cases, so Anubis uses its own branch (with a configurable name) instead. The last commit on Anubis `merge_branch` is the result of merging PR branch into PR target branch, as a regular `git merge` command would do. Anubis uses the _code tree_ from that last commit when creating a squashed commit on the `staging_branch`. The primary difference between the two branches is that `staging_branch` ends in a squashed (i.e. single-parent) commit rather than a regular two-parent merge commit; after both branches are updated, a `git diff staging_branch..bot_merge_branch` command produces no output because both branches contain the same code tree.

README.md
```diff
@@ -379,6 +379,7 @@ All configuration fields are required.
 *staged_run*| A boolean option to enable staging-only mode where the bot performs all the merging steps up to (but not including) the target branch update. Eligible PRs are merged into and tested on the staging branch but are never merged into their target branches. Staging-only mode prevents any target branch modifications by the bot. TODO: Check that the PR target branch is not the configured staging branch, setting `M-failed-other` if needed. | false
 *guarded_run*| Enables staging-only mode (see `config::staged_run`) for PRs without a `M-cleared-for-merge` label. Has no effect on PRs with that label. While `config::staged_run` blocks target branch modifications, this option allows them for a human-designated subset of PRs. | false
 *staging_branch* | The name of the bot-maintained git branch used for testing PR changes as if they were merged into their target branch. | auto
```
rousskov, line 379:
```md
* `tree`: The git tree object of the PR [merge
  commit](https://developer.github.com/v3/pulls/#get-a-single-pull-request)
  created by GitHub.
```

Please update the above text (located elsewhere in this README file) to say something like

```md
* `tree`: The git tree object from the last commit on the bot-maintained `bot_merge_branch`.
```

src/MergeContext.js
```diff
@@ -1721,17 +1726,15 @@ class PullRequest {
 
         assert(this._commitMessage.stageable);
 
+        // We have to create the merge commit manually because Github stopped generating merge commits
+        // reliably any time the base branch changes.
```
rousskov: Let's focus on what this code is doing and leave "why we had to do it" explanation to commit message and Anubis documentation:

```suggestion
        // To create a PR merge commit manually, 
        // we start by resetting our PR merge branch to become the same as PR base branch.
```

src/MergeContext.js
```diff
@@ -1721,17 +1726,15 @@ class PullRequest {
...
         const baseSha = await GH.getReference(this._prBaseBranchPath());
-        // We want to fast-forward this._mergeCommit code changes into the base branch, but we
-        // cannot use both this._mergeCommit.parents as this._stagedCommit parents because that
-        // would create a git merge commit, importing PR branch. We want flat history instead.
-        // Any commit created with baseSha as a parent can be fast-forwarded. To use baseSha, we must
-        // ensure that this._mergeCommit can still be fast-forwarded onto baseSha:
-        if (!this._mergeCommit.parents.some(p => p.sha === baseSha))
-            throw this._exLabeledFailure("PR merge commit is stale", Config.failedOtherLabel());
+        await GH.updateReference(Config.mergeBranchPath(), baseSha, true);
+        const mergeCommit = await GH.mergeBranch(Config.mergeBranch(), this._prHeadBranch());
```
rousskov: Let's add an empty line to separate `GH.updateReference()` (and its comment) above from the merging command below:

```suggestion

        // And then merge PR branch changes into our merge branch (as a two-parent merge commit).
        const mergeCommit = await GH.mergeAintoB(this._prHeadBranch(), Config.botMergeBranch());
```

The above suggestion also accounts for the "mergeAintoB()" change request that swaps the two parameters

src/MergeContext.js
```diff
@@ -1721,17 +1726,15 @@ class PullRequest {
...
+
         // If base branch changes after the above check, our _stagedPosition.ahead() checks
```
rousskov: That "above check" reference became stale with this PR changes:

```suggestion
        // We want to eventually fast-forward mergeCommit into the base branch, but we
        // cannot use both mergeCommit.parents as this._stagedCommit parents because that
        // would create a git merge commit, importing PR branch. We want flat history with one commit per PR instead.
        // If base branch changes after mergeCommit creation, our _stagedPosition.ahead() checks
```

src/MergeContext.js
```diff
@@ -1336,7 +1335,8 @@ class PullRequest {
 
     _createdAt() { return this._rawPr.created_at; }
 
-    _mergePath() { return "pull/" + this._rawPr.number + "/merge"; }
+    // Github-generated merge commit path
+    _defaultMergePath() { return "pull/" + this._rawPr.number + "/merge"; }
```
rousskov: The word "default" feels misleading here. It would also be good to flag this branch as often stale. I can suggest:


```suggestion
    // GitHub-generated PR merge branch. This branch may not reflect the latest target branch changes!
    _oldGhMergeBranchPath() { return "pull/" + this._rawPr.number + "/merge"; }
```

src/Config.js
```diff
@@ -90,6 +91,8 @@ class ConfigOptions {
     baseUrl() { return 'https://api.github.com'; }
     stagingBranchPath() { return "heads/" + this._stagingBranch; }
     stagingBranch() { return this._stagingBranch; }
+    mergeBranchPath() { return "heads/" + this._mergeBranch; }
+    mergeBranch() { return this._mergeBranch; }
```
rousskov: To better distinguish Anubis and GitHub merge branches:

```suggestion
    botMergeBranchPath() { return "heads/" + this._botMergeBranch; }
    botMergeBranch() { return this._botMergeBranch; }
```

Please propagate these renaming changes to all contexts.

src/Config.js
```diff
@@ -21,6 +21,7 @@ class ConfigOptions {
         this._port = conf.port;
         this._owner = conf.owner;
         this._stagingBranch = conf.staging_branch;
+        this._mergeBranch = conf.merge_branch;
```
rousskov: To better distinguish Anubis and GitHub merge branches:

```suggestion
        this._botMergeBranch = conf.bot_merge_branch;
```

Please propagate these renaming changes to all contexts.

src/MergeContext.js
```diff
@@ -1537,14 +1539,17 @@ class PullRequest {
...
+
+        const treeShaIsFresh = this._stagedCommit.tree.sha === mergeCommit.tree.sha;
         this._log("staged commit tree sha freshness: " + treeShaIsFresh);
         if (!treeShaIsFresh)
             return false;
 
         assert(this._stagedCommit.parents.length === 1);
         const stagedCommitParentSha = this._stagedCommit.parents[0].sha;
-        const parentIsFresh = this._mergeCommit.parents.some(p => p.sha === stagedCommitParentSha);
+        const parentIsFresh = mergeCommit.parents.some(p => p.sha === stagedCommitParentSha);
```
rousskov: We do not really _maintain_ the freshness of a bot-generated merge commit AFAICT:

* bot-generated merge commit may become stale compared to PR target branch;
* bot-generated merge commit may become stale compared to PR branch.

We care about the bot-generated merge commit in this context because it is the source of the staged code tree: If that source is stale, in any of the two "directions" above, then the staged code tree itself is stale. Please adjust this code to check all three things, in that order:

1. One of bot-generated merge commit parents matches the latest PR target branch SHA.
2. The other bot-generated merge commit parents matches the latest PR branch SHA.
3. The tree SHA of the bot-generated merge commit matches `this._stagedCommit.tree.sha`.

Finally, since the staged commit can (and sometimes is!) created by humans rather than Anubis, let's check that it has the right parent (in addition to the right code tree):

4. stagedCommitParentSha matches the latest PR target branch SHA.

I assume that the last check is not necessary to prevent merging of wrong staged commits with the right tree because our `GH.updateReference(...force:false)` merging would not allow that merging anyway, but it is best to detect the mismatch earlier, so that the bot can re-generate a problematic staged commit.

src/GitHubUtil.js
```diff
@@ -217,6 +217,20 @@ export async function updateReference(ref, sha, force) {
...
+    const result = await GitHub.rest.repos.merge(params);
+    if (result.status === 204) {
+        throw new Error(`did not merge ${head} into ${base}: already merged`);
+    }
+    assert(result.status === 201);
+    logApiResult(mergeBranch.name, params, {sha: result.data.sha});
```
rousskov: GitHub documents many outcomes that we should probably treat as errors, but none of those outcomes are clearly pointing to an Anubis code bug. Can we reorder/refactor this code to become something like


```suggestion
    logApiResult(mergeBranch.name, params, {sha: result.data.sha});
    if (result.status !== 201)
        throw new Error(`did not merge ${head} into ${base}: HTTP status code ${result.status}`);
```

rousskov (changes requested):
I think this PR is on the right track, thank you. Unfortunately, I believe we should rename and re-document a few things. I hope these partially overlapping change requests are not too confusing/conflicting.

src/Config.js
eduard-bagdasaryan: Ok.

src/Config.js
eduard-bagdasaryan: Ok.

src/GitHubUtil.js
eduard-bagdasaryan: Done.

src/GitHubUtil.js
eduard-bagdasaryan: Ok.

src/MergeContext.js

src/MergeContext.js

src/MergeContext.js
eduard-bagdasaryan: Ok, added these checks.

src/MergeContext.js

src/MergeContext.js

src/MergeContext.js

README.md
eduard-bagdasaryan: Applied with a minor adjustment.

README.md
eduard-bagdasaryan: Applied with a minor adjustment.

src/GitHubUtil.js
eduard-bagdasaryan: Ok.

eduard-bagdasaryan (commented):
Done (2296a1c).

README.md
```diff
@@ -248,9 +248,7 @@ formatting but strips any trailing whitespace.
 Anubis assembles the staging git commit (and, hence, in case of a successful
 PR merge, the target branch commit) from the following parts:
 
-* `tree`: The git tree object of the PR [merge
-  commit](https://developer.github.com/v3/pulls/#get-a-single-pull-request)
-  created by GitHub.
+* `tree`: The git tree object from the PR merge commit.
```
rousskov: That phrasing feels wrong/misleading because "PR merge commit" may mean "GitHub-maintained PR merge branch commit", and we are _not_ using that.


```suggestion
* `tree`: The git tree object from the last commit on the bot-maintained `bot_merge_branch`.
```

I know that you [rejected](https://github.com/measurement-factory/anubis/pull/81/changes#r3543783139) the above wording earlier, but I cannot guess why.

I am thinking that we can keep definitions on this list short and explain details where we define `bot_merge_branch`.

README.md
```diff
@@ -379,6 +377,7 @@ All configuration fields are required.
...
 *staging_branch* | The name of the bot-maintained git branch used for testing PR changes as if they were merged into their target branch. | auto
+*bot_merge_branch* | The name of the bot-maintained git branch that is very similar to GitHub-maintained PR merge branch. GitHub no longer updates its PR merge branches (i.e. refs/pull/PULL_REQUEST_NUMBER/merge references) in some important use cases, so Anubis uses its own branch (with a configurable name) instead. The last commit on Anubis bot_merge_branch (PR merge commit) is the result of merging PR branch into PR target branch, as a regular git merge command would do. Anubis uses the code tree from that last commit when creating a squashed commit on the staging_branch. The primary difference between the two branches is that staging_branch ends in a squashed (i.e. single-parent) commit rather than a regular two-parent merge commit; after both branches are updated, a git diff staging_branch..bot_merge_branch command produces no output because both branches contain the same code tree. | auto__
```
rousskov: Please use `backtics` for `refs/pull/PULL_REQUEST_NUMBER/merge`, `bot_merge_branch`, `staging_branch`, and `git diff staging_branch..bot_merge_branch`.

rousskov (changes requested):
Looks good to me, thank you! I only have two minor documentation polishing requests.

rousskov:
I also polished PR title and description a bit. Please check.

README.md
eduard-bagdasaryan: Done.

README.md
eduard-bagdasaryan: Adjusted with 'bot' and 'GitHub' prefixes.

eduard-bagdasaryan:
>  polished PR title and description

Checked.

rousskov (approved):

